Issue #40, opened 2019-07-09
[Bug] Can't save file in Dropbox folder


pezcode:
**Describe the bug**
Notepads doesn't save files inside my Dropbox folder. It opens and displays the content just fine, but pressing Save or CTRL+S doesn't remove the modified file icon, and the content isn't written to disk.
*Save as* actually creates the file, but it ends up empty.

**To Reproduce**
Not exactly tested. I have another laptop I'd love to test on but the Win 10 version is too old to install Notepads.

1. Install Dropbox (possibly on a non-system drive)
2. Create empty text file
3. Open with Notepads
4. Save
5. File is not saved

**Expected behavior**
The file should be saved. Neither Notepad, nor VS Code have trouble saving files in the same location.

**Desktop**
 - Windows 10 Pro 1903 (Version 10.0.18362)
 - Notepads v0.9.4.0

AmionSky:
It seems to be causing an exception here:
https://github.com/JasonStein/Notepads/blob/49b56d68bcce79d844f0d8cddb9c6ad11a609717/src/Notepads/MainPage.xaml.cs#L426

JasonStein:
I need to do some invesitigation here to better understand the handling logic for Dropbox, maybe Onedrive as well. I am currently rewritting/refactering the App logic, so I can only take a look at this later this week. Maybe @AmionSky can dig more into it? (Thanks for the prompt help by the way)

AmionSky:
Well I looked into it some more but was unable to find anything. It always returns an exception and as far as I looked online, no one really found a solution.

I'm not well versed in UWP so someone else might be able to figure out something.

Since it can still save the file if it doesn't use `DeferUpdates` and it does not seem that skipping it on exception can cause any problem, I updated my workaround PR with a slightly better way to handle this.

JasonStein:
Thanks @AmionSky 🎉 for the PR to address this issue. Code has been merged.

PR #43, opened 2019-07-09
Workaround for #40

AmionSky:
Workaround for #40

While it fixes the problem, I'm not sure if it won't cause incorrect behaviour when the `CachedFileManager.DeferUpdates` is needed.

Or maybe it should just `return false` in the catch so at least the app shows an error message.


JasonStein:
LGTM, l also did some research online, I think this is the way to go for now.

PR #48, opened 2019-07-12
Added de-DE and de-CH translation

WalterWolf49:
German translations for Germany (de-DE) and Switzerland (de-CH).


WalterWolf49:
Hi, I added the new translations.

PR #53, opened 2019-07-12
Fix unhandled exception on FileSavePicker

AmionSky:
If the FileSavePicker tricked into saving onto a read-only file, it throws an exception.

Also added a check so the app doesn't exit if a save failed

### DIFF
--- a/src/Notepads/MainPage.xaml.cs
+++ b/src/Notepads/MainPage.xaml.cs
@@ -639,7 +639,14 @@ namespace Notepads
                 !await FileSystemUtility.FileIsWritable(textEditor.EditingFile))
             {
                 NotepadsCore.SwitchTo(textEditor);
-                file = await FilePickerFactory.GetFileSavePicker(textEditor, _defaultNewFileName, saveAs).PickSaveFileAsync();
+                try
+                {
+                    file = await FilePickerFactory.GetFileSavePicker(textEditor, _defaultNewFileName, saveAs).PickSaveFileAsync();
+                }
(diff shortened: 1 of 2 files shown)


AmionSky:
When the FileSavePicker throws an exception, now it just returns false, but it does not notify the user that the save was unsuccessful.
Should it show a dialog or notification?

entdark:
> When the FileSavePicker throws an exception, now it just returns false, but it does not notify the user that the save was unsuccessful.
> Should it show a dialog or notification?

It definitely must show a dialog.

JasonStein:
One sec, I was not able to repro this because the picker prevents me from using read-only file and on the other hand if it fails, a new file save picker shows up. How did you test this case? What are the steps to repro?

AmionSky:
On the save file dialog select an already existing file.
Click on save so it shows the "Do you want to override?" message.
Change the file you about to override to read-only
Click yes on the override message.

JasonStein:
> 1. On the save file dialog select an already existing file.
> 2. Click on save so it shows the "Do you want to override?" message.
> 3. Change the file you about to override to read-only
> 4. Click yes on the override message.

Very smart...

JasonStein:
Forget about it, I am rewriting the save logic to matches open and produce exception. Also adding the string to resouce. Thanks! I think you can abandon this for now.

JasonStein:
https://github.com/JasonStein/Notepads/commit/c191cf6cf522bdd00e5fc0436df4fef4baf419b0

PR #54, opened 2019-07-12
Updated hu-HU translations

AmionSky:
Updated translations but I have a question.

The "AppCloseSaveReminderDialog_PrimaryButtonText" translated is too long to fit on it and I'm unsure how to fix it. "Save", "All" and "Exit" words aren't exactly the shortest in hungarian.

- Could the dialog be made wider/allow it to grow?
- Or the text on the buttons maybe omitting "Exit"?
- Or should I just say "Save & Exit"/"Save All" in hungarian?

image


JasonStein:
You do not have to do 1 on 1 mapping here, just make sure your word matches the behavior. I am perfectly fine with just "Save" and "Exit" here but you might want to check out how notepad.exe or vs code is doing the translation here.

AmionSky:
Already checked a few apps but it didn't help much. Updated it to "Save & Exit".

PR #82, opened 2019-07-23
Updated it-IT translation

guari:
Added translation for new strings


guari:
sorry for double commit, just made them shorter and aligned with other common apps syntax

PR #94, opened 2019-07-28
Updated hu-HU translation

AmionSky:


JasonStein:
two more here:

AmionSky:
Where are those strings located?

JasonStein:
https://github.com/JasonStein/Notepads/blob/c14dce99f0021cd4dcaeb0e0358a70db26867d4b/src/Notepads/Strings/hu-HU/Resources.resw#L149

AmionSky:
Looks like I wasn't on the right commit. Updated it now with the settings translations as well but I was unable to test it, VS complains about `No certificate found with the supplied thumbprint` when I try to build it.

JasonStein:
> Looks like I wasn't on the right commit. Updated it now with the settings translations as well but I was unable to test it, VS complains about `No certificate found with the supplied thumbprint` when I try to build it.

@validatedev is having the same issue with you. I need to check what went wrong here. I did not touch any build/config files recently to be clear. And I do not think certificate is required to build and run on dev box.

JasonStein:
I think you just need to clear cache, restart VS and rebuild.

PR #110, opened 2019-08-04
Font families filtering

Sergio0694:
✅ Filtered out some symbolic font families
✅ Sorted available fonts alphabetically


JasonStein:
LGTM

PR #111, opened 2019-08-04
Settings keyboard accelerator

Sergio0694:
Added (work in progress) `CTRL + ,` keyboard accelerator for the settings.
The code does look ok but for some reason it's not actually working, hence why it's a draft PR.
This might be related to the open issue related to this (linked in the code), though not 100% sure about this.

### DIFF
--- a/src/Notepads/MainPage.xaml.cs
+++ b/src/Notepads/MainPage.xaml.cs
@@ -76,6 +76,15 @@ namespace Notepads
         {
             InitializeComponent();
 
+            // Ctrl + . accelerator
+            KeyboardAccelerator accelerator = new KeyboardAccelerator
+            {
+                Key = (VirtualKey)0xBE, // 0xBE is mapped to VK_OEM_PERIOD
+                Modifiers = VirtualKeyModifiers.Control | VirtualKeyModifiers.Shift
+            };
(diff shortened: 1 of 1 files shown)


JasonStein:
KeyboardAccelerator works differently in 1809 and 1903, thus I am not using it in my App for now. All KeyboardAccelerators you can find in the app is only for showing the tooltip shortcut but disabled. I have code to handle all keyboard shortcuts which is more consistent and reliable.

JasonStein:
Btw, the settings for this app is not something user will frequently interact with. So I am not giving keyboard shortcut for it.

JasonStein:
But thanks for the invesitgation on it (appreciate!) and I would never know the issue you posted in the comment.

Sergio0694:
Ah, I see, makes sense. And yeah, I'm handling keyboard shortcuts manually as well in Legere, it's definitely more reliable. You might want to consider assigning that CTRL + , shortcut in the future though, it'd not be disruptive for the users as it's not a shurtcut that one would otherwise use anyway, and it'd be consistent with other apps like the new Windows Terminal or VS Code, that use the same shortcut for the settings as well.
I just think it'd be a nice little addition 😄

PR #114, opened 2019-08-08
Czech translate

JanRajnoha:


JasonStein:
Can you pull latest master? I added few more strings in resource.resw. Total line count should be 408.

JanRajnoha:
Added missing strings

PR #125, opened 2019-08-10
Updated cs-CZ translation

JanRajnoha:


JanRajnoha:
Updated cs-CZ translation

PR #150, opened 2019-08-16
Updated ja-JP Translation

pnp0a03:
I have a comment for original description.

I was confused at my first trial of this session backup/restore feature. With the description text "AdvancedPage_SessionBackupSettings_Description.Text", I understood the feature as following..."OK, my modified files will be `saved` automatically".
But, yes, this is not correct. The files are "Backup"ed within the notepads storage, but not "Saved" the original files.  To commit the changes to the original files, user should save it manually.
I believe that the description text should describe this behavior. In my translation, I'm using the following description (be patient for my broken English).

"Notepads will remember the current session for next launch and do a periodic make the temporary backup file to prevent accidental data loss. Notepads will NOT remind you to save your work on app close. And, open the temporary backup at next launch. To commit your changes to the original file, you should do the save action."


JasonStein:
I should name it session snapshot instead of backup and restore. It is more accurate, maybe something like these:

"Notepads will remember the current session for next launch and do a periodic session snapshot backup to prevent accidental data loss for uncommitted changes. Notepads will NOT remind you to save your work on app close if this feature is enabled"

JasonStein:
https://github.com/JasonStein/Notepads/commit/ae8d8014710e0f2709251bd43edf3299afb985ce

pnp0a03:
I like the new wording "Session snapshot". :) It's good.
I've changed the ja-JP translation based on the new wording.

PR #154, opened 2019-08-18
pt-BR translation

pinholucas:


JasonStein:
Since there is a code change and it has to be adjusted for all languages, I will merge it later once I got a free chunk of time.

pinholucas:
Ok, tell me if I can help with anything about it. :)

PR #165, opened 2019-08-30
Added close button

kephax:
Added close button, similar to the classic notepad


JasonStein:
There are too many buttons in the menu so I never intended to add Exit button. On the other hand, once last tab is closed, the app will exit automatically so there is no need to have this button. User should always use system close button to exit app. This is recommended.

JasonStein:
If you want to add a feature, please create one in issues first and discuss with me and others.

JasonStein:
Btw,                                             <MenuFlyoutItem.KeyboardAccelerators>
                                                <KeyboardAccelerator Key="Q" IsEnabled="True"/>
                                            </MenuFlyoutItem.KeyboardAccelerators>

This means a simple "Q" will terminate the app, this should not be the case.

PR #166, opened 2019-08-30
Added ctrl + D to duplicate line or selection

pinholucas:
It works with the current line and selection, but isn't working yet with wrapped lines.

When you try to duplicate a selection, starting from the end of a line, it selects a imaginary (? ahaha) space. I don't know the reason, probably something related to the line endings.

Notepads
image

Notepad
image


JasonStein:
Can you target master branch instead of my feature branch?

pinholucas:
> Can you target master branch instead of my feature branch?

Ok, I think that I need to make a new pull request, I'm right?

JasonStein:
> > Can you target master branch instead of my feature branch?
> 
> Ok, I think that I need to make a new pull request, I'm right?

Yes

PR #167, opened 2019-08-31
Added ctrl + D to duplicate current line or selection

pinholucas:
It works with the current line and selection, but isn't working yet with wrapped lines.

When you try to duplicate a selection, starting from the end of a line, it selects a imaginary (? ahaha) space. I don't know the reason, probably something related to the line endings.

Notepads
image

Notepad
image


pinholucas:
Is it right now? I'm still a newbie with git ahahhaha 😅

JasonStein:
I tried with your changes, for the line insertion, it works perfectly. It is a little bit confusing for selection duplication tho since you are not excluding the line ending. For line ending, you can always assume '\n' here since other type of line ending will be converted to '\n' by built-in RichEdit control.

On the other hand, to make it work with wrapped line, you need to actually calculate line position, selection position yourself instead of using TextRange built-in methods. One example you can look at is here: https://github.com/JasonStein/Notepads/blob/ebe199e01e04c5204b3a4ec39f2e9504f9c78941/src/Notepads/Controls/TextEditor/TextEditorCore.cs#L164

pinholucas:
Ok, I'll do it tomorrow

pinholucas:
After you modified the GetCurrentLineColumn method, adding the GetTextSelectionPosition, my code isn't working 100% anymore. But works again when I change the GetTextSelectionPosition to the old:
var start = Document.Selection.StartPosition;
var end = Document.Selection.EndPosition;

JasonStein:
> After you modified the GetCurrentLineColumn method, adding the GetTextSelectionPosition, my code isn't working 100% anymore. But works again when I change the GetTextSelectionPosition to the old:
> var start = Document.Selection.StartPosition;
> var end = Document.Selection.EndPosition;

You are right, the new GetTextSelectionPosition() is used as a thread-safe approach to get current position data in a background thread for session recording. It won't be updated in time with synchronized code, while the old way will work:
> var start = Document.Selection.StartPosition;
> var end = Document.Selection.EndPosition;
since they will return positions from UI thread and will always be accurate.

JasonStein:
I just played with your code, I noticed that you are not following existing behavior found in Notepad++ and Visual Studio, can you make it feature parity with Notepad++ and Visual Studio? So it makes more sense to users. Like ctrl+d on selected text, it should paste it right after the cursor. And ctrl+d at first character of the line should not create a new empty line. If you do not have time, I can take over from here as well. You decide.

pinholucas:
No problem, I'll do it later today.

pinholucas:
"today + 5" 😅
Sorry for the delay, this week I've had some problems in life to solve.

By my tests, now it's working like the ctrl + E, V of Visual Studio. If there's any problem, tell me :)

JasonStein:
> "today + 5" 😅
> Sorry for the delay, this week I've had some problems in life to solve.
> 
> By my tests, now it's working like the ctrl + E, V of Visual Studio. If there's any problem, tell me :)

No worries, you take your time. I will take a look later and give you suggestions if any. Thanks for doing it! 💯

JasonStein:
It works great, I have not found any issue yet. I will test with edge cases and merge. LGTM,

PR #169, opened 2019-08-31
2 new buttons

FJduFou:
contributors + downloads

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,9 @@
 [![Beta release](https://img.shields.io/github/release/jasonstein/notepads.svg?label=beta%20release&style=flat-square)](https://github.com/JasonStein/Notepads/releases)
 ![Platform](https://img.shields.io/badge/platform-windows%2010%20%7C%20uwp-yellow.svg?style=flat-square)
 [![Discord](https://img.shields.io/discord/588473626651787274.svg?style=flat-square)](https://discord.gg/VqetCub)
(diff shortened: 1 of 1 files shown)


JasonStein:
I used to have the download link but it is not suitable for now since all users are going to Microsoft Store for app downloading of Notepads. There are not many traffic from release page now. 

For contributor, there is already a dedicated section, so no button is needed as well.

The most important reason is that there are already 4 badges there, 5 would be too much and no one will every care. Just my 2 cents here.

FJduFou:
for the contributor, it is to have an overview
otherwise, for the download button, it is a question of seeing if it is a very downloaded application or not
especially since yours is very downloaded, it can perhaps encourage developers to participate in the project ...

JasonStein:
Only few download notepads from the release page, the number won't be appealing. The star count is good enough for attracting developers, btw, I still prefer solo at this moment since I am still passionate about what I am doing right now. I would prefer to have a list of contributor icons down below readme.

FJduFou:
Where exactly the contributors ?

JasonStein:
I am referring to something like these:
image

FJduFou:
ok

FJduFou:
https://sourcerer.io
if you register in this, i think is possible

JasonStein:
I have added hall-of-fame in the contributor section and tweak a little bit on readme.

FJduFou:
okki

PR #184, opened 2019-09-12
Updated pt-BR translation

pinholucas:
I don't know if its a problem with my windows or with the code, but I couldn't open a shadow instance of the app (shift + left click, if I'm not wrong), also couldn't make the "move tab here" appear.


JasonStein:
There is a bug around opening multi-instances and I have fixed it. So for now, you can only open a new (shadow) instance by dragging out a new document tab. I will add a jump-list action later.

JasonStein:
or you can open "run" and execute: notepads://newinstance

image

PR #185, opened 2019-09-12
Added Go To Line 

pinholucas:
It has the same behavior than VS Code (not close the popup when it can't find the line), and the same layout as the Find/Replace popup. Also the icon that I used, is basically the same as the one used in Office.

For the errors, it's using the Find/Replace error message, what I think is fine.

Let me now if there's anything wrong. 😄


pinholucas:
Done. :)

I would suggest two things that I'm not sure in how to do it: Change the background color when mouse over the status bar 'buttons' and close the GoTo popup when it lose focus.

JasonStein:
I have some immediate items to finish and I will come back to your PR at a later time. Sounds good?

pinholucas:
If you agree with the suggestions that I've made, I can work on it later.

JasonStein:
> If you agree with the suggestions that I've made, I can work on it later.

Can you take a look at "changing background color on mouse hover"? This is something I wanted to do as well. For line ending, encoding, line colum indicators you can change background opacity to 20% for the grid or textbox. And for name/path indicator you can add underline for the path on mouse hover. There should be an easy way to do it without code change, consider using this: https://github.com/microsoft/XamlBehaviors

pinholucas:
Hey, is it supposed to be working? I can't make it appear while running the app and it seems that has the same functionality as the "ModificationIndicator" at column 2.

image

pinholucas:
Done! 

- I've removed the minWidth='10' from the first status bar column, because I can't find a use for it, maybe you forgot to remove? Tell me if I'm wrong.
- All the status bar elements acts like a button, when the mouse is over them. It's the behavior of VS Code, I think that is better to copy it from them.

JasonStein:
> Done!
> 
> * I've removed the minWidth='10' from the first status bar column, because I can't find a use for it, maybe you forgot to remove? Tell me if I'm wrong.
> * All the status bar elements acts like a button, when the mouse is over them. It's the behavior of VS Code, I think that is better to copy it from them.
> * Now the GoTo dialog works like in VS Code, it closes when the TextBox loses focus, and I've removed the close button from the previous commit.

Thanks! I will take a deep look this weekend and provide feedback.

pinholucas:
What about adding tooltips to the statusbar buttons, like in VS Code?

image

JasonStein:
No tooltips for now, I will add that part later.

And now I remember why I set min-width there:

image

JasonStein:
While, in the meantime, if you are bored and like to do something for Notepads, you can PM me and I will see what items on the plate that I might need help from you. Thanks for the continues contribution btw!

PR #209, opened 2019-09-28
Recording & restoration

FJduFou:
Modify for a good size
+ fix 2 small mistakes


JasonStein:
please use "&amp;" for "&"
the build is broken
https://stackoverflow.com/questions/9084237/what-is-amp-used-for

PR #216, opened 2019-10-01
Fix not showing full file content when there is a NULL character in the middle of the file

### LINKED ISSUE #200: [Bug] Cannot read the Bitlocker recovery key text file properly
**Describe the bug**
As you see on the screenshot, the Bitlocker encrypt recovery key file cannot be seen properly with Notepads, it works flawlessly on Notepad.

**To Reproduce**
Create the Bitlocker encryption process and save the recovery key on the txt file.

**Expected behavior**
The recovery key text file content should be visible correctly with the Notepads.

**Screenshots**
bug

**Desktop (please complete the following information):**
 - OS: Windows 10 1903 18362.356
 - Version v0.10.2.0

validatedev:
Fixes #200 
Co-authored-by: M. Şamil Ateşoğlu <m.samilatesoglu@gmail.com>


JasonStein:
I am not sure if we can simply remove it from the string. That means, when we save the file, the null character will be gone right?

JasonStein:
Take a look at last comment in this thread: https://www.c-sharpcorner.com/forums/how-to-write-null-character-in-textbox

validatedev:
I think that the problem is the bug which is located on RichEditBox. It should display the file until encountering the EOF. Let me explain with details:

`text = reader.ReadToEnd();`

That implementation of reading the file is true and works as intended. It reads and passes the "\0" character, not stops there. But the built-in tool on the Visual Studio that shows the content what the variable has encounters the same error, it doesn't show the file content after the "\0" character. FileInput and string are not the same thing. The built-in tool and the RichEditBox also behave to file like a just the string. Let me explain with photos:

_Before the changes_



_After the changes_



And good news, our implementation works correctly without any problem. It doesn't change the string "\0", just the character one. Also you can try to change RichEditBox thing to display characters correctly but it's just up to you.

JasonStein:
Yes, I know it works from reading to representation. I am more concerned about file saving. Did you try open the file and save it? Then take a look if \0 is missing in the output file?

validatedev:
> Yes, I know it works from reading to representation. I am more concerned about file saving. Did you try open the file and save it? Then take a look if \0 is missing in the output file?

It is gone from the file but I don't think that is the problem. It can surpass the errors that may happen in the future.

JasonStein:
> > Yes, I know it works from reading to representation. I am more concerned about file saving. Did you try open the file and save it? Then take a look if \0 is missing in the output file?
> 
> It is gone from the file but I don't think that is the problem. Also it can surpass the errors that may happen in the future.
> 
> And one last note: If you don't want to delete \0, you should look at the behaviour of the RichEditBox. If you can change that thing with their superior, it will definitely more accurate solution than this implementation but I can't see any harm with that.

No, actually it is a problem, because if the file is intended to have "\0" and it is accepted by another program, then it will cause unwanted behavior. Let me think about it and see what we can do here.

JasonStein:
At least, we can(should) let the user know if the file opening has a null character and let the user decide what to do (ignore or remove it).

validatedev:
Ok then I'll close that pull req and try to apply the solution that you're referring to.

PR #217, opened 2019-10-05
Add warning about NULL character on the file

validatedev:


validatedev:
Temporary fix of #200

validatedev:
If you really want to remove the null chars on the file button on warning dialog, I’ll try to do it but that solution gives more sense then you said to me.

itsusmanmirza:
Could you please add the auto-save feature like saving a file every few seconds I usually use CTRL+S to save it but still sometimes when in a hurry, I forget to save it.

validatedev:
@JasonStein could you take a look at it, saying that as a precaution about that I don't know if you see it.

JasonStein:
This is the best place to call the dialog and the method should returns error and handled in main view. Other than that I still prefer not to have it at all but warn user instead. Something like "your file has null character, only content before the null character is showing" and it should be a notification instead of dialog and we do nothing in this case. Once we gather feedback from user, as well as metrics event. We can decide what to do next. Let me think about it and I can do it myself. Still, thx for the PR!

PR #266, opened 2019-12-30
Update hu-HU translations

AmionSky:
Didn't test it because I don't have VS installed right now, but should be ok.


FJduFou:
it's good

JasonStein:
Thanks!

PR #282, opened 2020-01-26
file icons

mah-qurashy:
custom file icons for most supported files


JasonStein:
Thanks! This is amazing:)

PR #290, opened 2020-02-09
Updated cs-cz

JanRajnoha:


JasonStein:
Thanks!

JanRajnoha:
I had many exams and other tasks with higher priority, so I'm sorry for little delay.

PR #293, opened 2020-02-13
Added some features

soumyamahunt:
Fully implemented go to line control with removing closing on losing focus, added limit to only allow number input and number should be less than number of lines.
Added option to highlight current line.
Added feature to do web search also provided option to change search engine.
Added date/time input to context menu for better discoverability.
Added new item to statusbar to show percentage zoom in editor.
Modified text zooming to increase/decrease by 10%.
Added flyout to line indicator for quick access to word wrap, highlight current line and go to line control.


JasonStein:
Superb! I will review and provide feedback. Thanks for the big contribution here.

src/Notepads/Controls/TextEditor/TextEditor.xaml.cs
```diff
@@ -889,5 +952,47 @@ private void GoToControl_OnDismissKeyDown(object sender, RoutedEventArgs e)
             GoToPlaceholder.Dismiss();
             TextEditorCore.Focus(FocusState.Programmatic);
         }
+
+        private void DrawLineHighlighter()
+        {
+            TextEditorCore.Document.Selection.GetRect(Windows.UI.Text.PointOptions.ClientCoordinates, out Windows.Foundation.Rect highlightRect, out var _);
+            LineHighlighter.Height = 1.35 * TextEditorCore.FontSize;
```
JasonStein: How did you come up with this magic number? (1.35)

soumyamahunt: On trying with different font sizes and different font factors, this is the one that worked perfectly.

soumyamahunt:
> Superb! I will review and provide feedback. Thanks for the big contribution here.

Glad to contribute, thanks for starting this project.

src/Notepads/Controls/TextEditor/TextEditor.xaml.cs
JasonStein: That explains. I have few more suggestions regarding to your PR and there are some places need to be polished. I can merge your PR right away and do it on my own or I can leave you comments. Which way do you prefer?

soumyamahunt: Leave me comments and I will do it right away.

JasonStein: 1. Bottom edge of the highlighter is way too close to the bottom of the line. Increase it a little bit would make it better.
image

2. I do not think adding "Wrap Word" here is necessary since it is already accessible in right click context menu. As well for "Highlight Current Line", it is not something that user will change on a daily basis. So keeping it in the settings is good enough. Basically you can remove this menu completely and whenever user click on line number indicator, the GoTo should pop up.
image

3. Remove the hint "Restore Default Zoom (Ctrl+0)" and remove the gap between four edges:
image

4. Add a GoTo Text string right before the text box to make it clear. Otherwise user will be confused and will think it is the normal search bar:
image

5. Make this feature enabled by default since there is no harm of doing so and most of the users won't know it if they do not go to settings on a daily basis:
image

6. Increase the top and bottom gap of the search engine custom textbox will make it look better:
image

src/Notepads/Controls/Settings/TextAndEditorSettings.xaml.cs
```diff
@@ -60,6 +60,7 @@ public TextAndEditorSettings()
 
             TextWrappingToggle.IsOn = (EditorSettingsService.EditorDefaultTextWrapping == TextWrapping.Wrap);
             HighlightMisspelledWordsToggle.IsOn = EditorSettingsService.IsHighlightMisspelledWordsEnabled;
+            LineHighlighterToggle.IsOn = EditorSettingsService.EditorDefaultLineHighlighterViewState;
```
JasonStein: EditorDefaultLineHighlighterViewState is good but let's keep our existing naming convention here by renaming it to "IsLineHighlighterEnabled"

src/Notepads/Controls/TextEditor/TextEditor.xaml.cs
```diff
@@ -186,6 +189,14 @@ public TextEditor()
             base.Loaded += TextEditor_Loaded;
             base.Unloaded += TextEditor_Unloaded;
             base.KeyDown += TextEditor_KeyDown;
+
+            base.SizeChanged += (s, e) => ToggleLineHighlighterView(EditorSettingsService.EditorDefaultLineHighlighterViewState);
+            TextEditorCore.ScrollViewerOffsetChanged += (s, e) => ToggleLineHighlighterView(EditorSettingsService.EditorDefaultLineHighlighterViewState);
```
JasonStein: Move all "(s, e) =>" into separate method.

src/Notepads/Controls/TextEditor/TextEditor.xaml.cs
```diff
@@ -186,6 +189,14 @@ public TextEditor()
...
+            EditorSettingsService.OnDefaultLineHighlighterViewStateChanged += (sender, visibility) => ToggleLineHighlighterView(visibility);
```
JasonStein: Same for all these.

src/Notepads/Controls/TextEditor/TextEditor.xaml.cs
```diff
@@ -217,6 +228,14 @@ public void Dispose()
             base.Unloaded -= TextEditor_Unloaded;
             base.KeyDown -= TextEditor_KeyDown;
 
+            base.SizeChanged -= (s, e) => ToggleLineHighlighterView(EditorSettingsService.EditorDefaultLineHighlighterViewState);
+            TextEditorCore.ScrollViewerOffsetChanged -= (s, e) => ToggleLineHighlighterView(EditorSettingsService.EditorDefaultLineHighlighterViewState);
```
JasonStein: And these as well.

src/Notepads/Controls/TextEditor/TextEditor.xaml.cs
```diff
@@ -889,5 +952,47 @@ private void GoToControl_OnDismissKeyDown(object sender, RoutedEventArgs e)
             GoToPlaceholder.Dismiss();
             TextEditorCore.Focus(FocusState.Programmatic);
         }
+
+        private void DrawLineHighlighter()
+        {
+            TextEditorCore.Document.Selection.GetRect(Windows.UI.Text.PointOptions.ClientCoordinates, out Windows.Foundation.Rect highlightRect, out var _);
+            LineHighlighter.Height = 1.35 * TextEditorCore.FontSize;
+
+            Thickness LineHighlighterBorderThickness = new Thickness(0.1 * LineHighlighter.Height);
```
JasonStein: Create a const var for all "0.1" references in this method.

src/Notepads/Controls/TextEditor/TextEditorCore.cs
```diff
@@ -481,13 +497,13 @@ private void SetDefaultTabStopAndLineSpacing(FontFamily font, double fontSize)
...
         {
-            FontSize += delta;
+            FontSize = FontSize + delta * EditorSettingsService.EditorFontSize;
         }
 
         private void DecreaseFontSize(double delta)
         {
-            if (FontSize < delta + 2) return;
-            FontSize -= delta;
+            if (FontSize < 0.11 * EditorSettingsService.EditorFontSize) return;
```
JasonStein: what is this 0.11?

src/Notepads/Services/EditorSettingsService.cs
```diff
@@ -276,6 +319,18 @@ private static void InitializeSpellingSettings()
...
 
+        private static void InitializeLineHighlighterSettings()
+        {
+            if (ApplicationSettingsStore.Read(SettingsKey.EditorDefaultLineHighlighterViewStateStr) is bool lineHighlighterStr)
+            {
+                _editorDefaultLineHighlighterViewState = lineHighlighterStr;
+            }
+            else
+            {
+                _editorDefaultLineHighlighterViewState = false;
```
JasonStein: Make it "true" by default.

src/Notepads/Utilities/SearchEngineUtility.cs
```diff
@@ -0,0 +1,39 @@
...
+
+    public enum SearchEngine
+    {
+        Bing,
+        Google,
+        DuckDuckGo,
+        Custom
+    }
+
+    class SearchEngineUtility
```
JasonStein: You can just create a static Dictionary<enum, string> here to make it simple.

src/Notepads/Controls/TextEditor/TextEditor.xaml.cs
soumyamahunt: About changing the linehighlighter height, I chose 1.35 cause the selection height is also the same. 
ApplicationFrameHost_2020-02-15_18-08-01
For my research into other text editors that implement the same behaviour, the bottom of the line is also a lot closer. I have no problem increasing the height but it looks weird when words are selected. 
ApplicationFrameHost_2020-02-15_18-05-20
Shall I disable highlight when selection length is greater than zero?

src/Notepads/Controls/TextEditor/TextEditorCore.cs
soumyamahunt: I wanted the font size to only decrease when the size is greater or equal to 10% of the font size chosen in settings.

src/Notepads/Controls/TextEditor/TextEditor.xaml.cs
JasonStein: Ok, makes sense. Let's keep what it is. We can keep the higlight even if selection length is greater than zero.

soumyamahunt:
All the changes suggested are done, and it's ready to be merged.

JasonStein:
> All the changes suggested are done, and it's ready to be merged.

One more bug I found:
image

When "Word Wrap" is enabled and :
1. First, click on second line.
2. Then click on last character or end of the line of first line.
The highlighter will not change it's postion. Instead it should highlight first line.

soumyamahunt:
> When "Word Wrap" is enabled and :
> 
> 1. First, click on second line.
> 2. Then click on last character or end of the line of first line.
>    The highlighter will not change it's postion. Instead it should highlight first line.

This seems to be a bug in Richeditbox control, in the above case clicking far from the end of the first part of the line returns the properties for the leftmost position of the next part of the line(you can see that in line column indicator). This only happens when selection is changed by click only ,there is no way to go to that position with arrow keys.
ApplicationFrameHost_2020-02-16_19-07-54
ApplicationFrameHost_2020-02-16_19-12-14
I can't seem to find a way to resolve this.

> There is no upper limit of doing "Ctrl+mouse wheel up" or "Ctrl+'+'":
> Notice, it shows 630% instead of 500%. Maybe we should cap it to 500% and the IncreaseFontSize method should only increase font size when it is under 500% zooming.

I wanted users to be able to change font size without fondling into settings too much and in case of shadow windows the only way to do so is by "Ctrl+mouse wheel up" or "Ctrl+'+'". In my opinion instead of restricting increase limit showing showing >500% or the actual zooming %  would be a good idea. Let me know your opinion.

JasonStein:
> > When "Word Wrap" is enabled and :
> > 
> > 1. First, click on second line.
> > 2. Then click on last character or end of the line of first line.
> >    The highlighter will not change it's postion. Instead it should highlight first line.
> 
> This seems to be a bug in Richeditbox control, in the above case clicking far from the end of the first part of the line returns the properties for the leftmost position of the next part of the line(you can see that in line column indicator). This only happens when selection is changed by click only ,there is no way to go to that position with arrow keys.
> ApplicationFrameHost_2020-02-16_19-07-54
> ApplicationFrameHost_2020-02-16_19-12-14
> I can't seem to find a way to resolve this.
> 
> > There is no upper limit of doing "Ctrl+mouse wheel up" or "Ctrl+'+'":
> > Notice, it shows 630% instead of 500%. Maybe we should cap it to 500% and the IncreaseFontSize method should only increase font size when it is under 500% zooming.
> 
> I wanted users to be able to change font size without fondling into settings too much and in case of shadow windows the only way to do so is by "Ctrl+mouse wheel up" or "Ctrl+'+'". In my opinion instead of restricting increase limit showing showing >500% or the actual zooming % would be a good idea. Let me know your opinion.

Ok, sounds good to me. Let's keep what it is. The built-in UWP RichEdit is really annoying...

src/Notepads/App.xaml.cs
```diff
@@ -217,6 +217,15 @@ private async System.Threading.Tasks.Task ActivateAsync(IActivatedEventArgs e)
...
                     "IsHighlightMisspelledWordsEnabled", EditorSettingsService.IsHighlightMisspelledWordsEnabled.ToString()
+                },
+                {
+                    "IsLineHighlighterEnabled", EditorSettingsService.IsLineHighlighterEnabled.ToString()
+                },
+                {
+                    "EditorDefaultSearchEngine", EditorSettingsService.EditorDefaultSearchEngine.ToString()
+                },
+                {
+                    "EditorCustomMadeSearchUrl", EditorSettingsService.EditorCustomMadeSearchUrl.ToString()
```
JasonStein: Remove this line since it might expose user privacy related information.

src/Notepads/Controls/GoTo/GoToControl.xaml.cs
```diff
@@ -14,6 +15,18 @@ public sealed partial class GoToControl : UserControl
 
         public event EventHandler<GoToEventArgs> OnGoToButtonClicked;
 
+        private int CurrentLine { get; set; }
```
JasonStein: _currentLine

src/Notepads/Controls/GoTo/GoToControl.xaml.cs
```diff
@@ -14,6 +15,18 @@ public sealed partial class GoToControl : UserControl
...
+
+        private int MaxLine { get; set; }
```
JasonStein: _maxLine

src/Notepads/Controls/TextEditor/TextEditor.xaml.cs
```diff
@@ -889,5 +952,77 @@ private void GoToControl_OnDismissKeyDown(object sender, RoutedEventArgs e)
...
+                    LineHighlighter.BorderThickness = new Thickness(lineHighlighterBorderThickness, 0, lineHighlighterBorderThickness, lineHighlighterBorderThickness);
+                    LineHighlighter.Height += lineHighlighterMargin.Top;
+                    LineHighlighter.Visibility = Visibility.Visible;
+                }
+                else
+                    LineHighlighter.Visibility = Visibility.Collapsed;
+            }
+        }
+
+        private void LineHighlighter_OnViewStateChanged(object sender, bool e)
```
JasonStein: e => enabled

JasonStein:
Also, feel free to add your name here in your PR:

image

soumyamahunt:
> When "Word Wrap" is enabled and :
> 
> 1. First, click on second line.
> 2. Then click on last character or end of the line of first line.
>    The highlighter will not change it's postion. Instead it should highlight first line.

I have reported the bug in WinUI repo [here](https://github.com/microsoft/microsoft-ui-xaml/issues/1988).
Also I found a new bug which is related to RichEditBox and reported it [here](https://github.com/microsoft/microsoft-ui-xaml/issues/1989).

src/Notepads/NotepadsMainPage.StatusBar.cs
```diff
@@ -195,6 +208,63 @@ private async void ReloadFileFromDisk(object sender, RoutedEventArgs e)
...
+        }
+
+        private void FontZoomSlider_ValueChanged(object sender, RangeBaseValueChangedEventArgs e)
+        {
+            if (!(sender is Slider item)) return;
+
+            var selectedTextEditor = NotepadsCore.GetSelectedTextEditor();
+            if (selectedTextEditor == null) return;
+
+            if (e.NewValue != e.OldValue)
```
JasonStein: This will never be false since range value is a double.
You can round them into integer and compare or write something like this:
if (Math.Abs(e.NewValue - e.OldValue) > TOLERANCE)

src/Notepads/Services/EditorSettingsService.cs
```diff
@@ -276,6 +319,18 @@ private static void InitializeSpellingSettings()
             }
         }
 
+        private static void InitializeLineHighlighterSettings()
+        {
+            if (ApplicationSettingsStore.Read(SettingsKey.EditorDefaultLineHighlighterViewStateBool) is bool lineHighlighterStr)
```
JasonStein: lineHighlighterStr => isLineHighlighterEnabled

src/Notepads/Services/EditorSettingsService.cs
```diff
@@ -342,6 +397,21 @@ private static void InitializeTabIndentsSettings()
             }
         }
 
+        private static void InitializeSearchEngineSettings()
+        {
+            if (ApplicationSettingsStore.Read(SettingsKey.EditorDefaultSearchEngineStr) is string searchEngineStr && ApplicationSettingsStore.Read(SettingsKey.EditorCustomMadeSearchUrlStr) is string customMadesearchUrl)
+            {
+                Enum.TryParse(typeof(SearchEngine), searchEngineStr, out var searchEngine);
```
JasonStein: If failed, set search engine to Bing.

src/Notepads/Utilities/SearchEngineUtility.cs
```diff
@@ -0,0 +1,29 @@
...
```
JasonStein: Make it public static

src/Notepads/Utilities/SearchEngineUtility.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        private static Dictionary<SearchEngine, string> SearchEngineUrlDictionary = new Dictionary<SearchEngine, string>
```
JasonStein: make it private static "readonly"

JasonStein:
Left few more comments. Please take a look. Thanks!

src/Notepads/Controls/TextEditor/TextEditor.xaml.cs
```diff
@@ -854,6 +914,9 @@ public void ShowGoToControl()
             if (GoToPlaceholder.Visibility == Visibility.Collapsed)
                 GoToPlaceholder.Show();
 
+            GetCurrentLineColumn(out var line, out var column, out var selectedLength);
+            var maxLine = TextEditorCore.GetText().Length - TextEditorCore.GetText().Replace("\r", string.Empty).Length + 1;
```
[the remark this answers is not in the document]
soumyamahunt: Text wrapping doesn't affect it in any way, did you come accross any bug related to this.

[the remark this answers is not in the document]
JasonStein: You are right, please ignore me.

JasonStein:
LGTM, ready for merge?

soumyamahunt:
> LGTM, ready for merge?

Yes, go right ahead.

JasonStein:
> > LGTM, ready for merge?
> 
> Yes, go right ahead.

It's merged but there are some bugs and issues with your changes. I am going to fix it in the next commit.

1. Selection calculation is wrong comparing to Visual Studio Code. I will revert my previous calculation code.
2. Set indicator gray out feature will not work properly in some scenarios like window resizing.
3. I am not sure if DuplicateText() method you touched works as before, can you verify?
4. GoTo bar LostFocus="GoToBar_LostFocus" is missing.

soumyamahunt:
> It's merged but there are some bugs and issues with your changes. I am going to fix it in the next commit.
> 
> 1. Selection calculation is wrong comparing to Visual Studio Code. I will revert my previous calculation code.
> 2. Set indicator gray out feature will not work properly in some scenarios like window resizing.
> 3. I am not sure if DuplicateText() method you touched works as before, can you verify?
> 4. GoTo bar LostFocus="GoToBar_LostFocus" is missing.
1. The selection calculation I implemented rejects line ending characters but vs code calculates it. I can modify the method to show like vs code.
2. I have  fixed the grey out issue.
3. DuplicateText works flawlessly in all test cases I tried.
4. I provided dismiss button for gotobar so the lost focus method not needed.

JasonStein:
> > It's merged but there are some bugs and issues with your changes. I am going to fix it in the next commit.
> > 
> > 1. Selection calculation is wrong comparing to Visual Studio Code. I will revert my previous calculation code.
> > 2. Set indicator gray out feature will not work properly in some scenarios like window resizing.
> > 3. I am not sure if DuplicateText() method you touched works as before, can you verify?
> > 4. GoTo bar LostFocus="GoToBar_LostFocus" is missing.
> 
> 1. The selection calculation I implemented rejects line ending characters but vs code calculates it. I can modify the method to show like vs code.
> 2. I have  fixed the grey out issue.
> 3. DuplicateText works flawlessly in all test cases I tried.
> 4. I provided dismiss button for gotobar so the lost focus method not needed.

1. I think my previous method is more efficient and do not require recalculation if not needed and no string manipulation is needed. So we should keep it.
4. If you have GoTo bar opened and add or delete lines. The GoTo bar will not update its current line and max, which will cause problem.  LostFocus is good here.

soumyamahunt:
> 1. I think my previous method is more efficient and do not require recalculation if not needed and no string manipulation is needed. So we should keep it.
> 2. If you have GoTo bar opened and add or delete lines. The GoTo bar will not update its current line and max, which will cause problem.  LostFocus is good here.

1. For large number of lines for loop might not be efficient, Also storing all the split lines might cause overhead. I have modified the method to act exactly like VS Code. You can choose to keep either one.
2. The problem with lost focus was when clicked on the hamburger menu the flyout disappears along with the gotocontrol. I have modified the method to dismiss gotocontol when text changed.

JasonStein:
@soumyamahunt I am receiving unhandled exception from some of the users:

Exception
Windows.UI.Xaml.Markup.XamlParseException: XAML parsing failed.	
Cannot find a Resource with the Name/Key AppBarButtonContentHeight"

OSVersion
10.0.17763.914

Can you help take a look? 
It is related to CustomAppBarButtonLabelToRightStyle:

image

soumyamahunt:
> @soumyamahunt I am receiving unhandled exception from some of the users:
> 
> Exception
> Windows.UI.Xaml.Markup.XamlParseException: XAML parsing failed.
> Cannot find a Resource with the Name/Key AppBarButtonContentHeight"
> 
> OSVersion
> 10.0.17763.914
> 
> Can you help take a look?
> It is related to CustomAppBarButtonLabelToRightStyle:
> 
> image

I used [this](https://stackoverflow.com/questions/35937713/xaml-parsing-failed-visual-studio-2013-windows-ui-xaml-markup-xamlparseexcepti) article to create the custom appbar button. I couldn't reproduce the issue however I am removing the key since it isn't affecting the design in any way. Will submit a pull request soon.

JasonStein:
> > @soumyamahunt I am receiving unhandled exception from some of the users:
> > Exception
> > Windows.UI.Xaml.Markup.XamlParseException: XAML parsing failed.
> > Cannot find a Resource with the Name/Key AppBarButtonContentHeight"
> > OSVersion
> > 10.0.17763.914
> > Can you help take a look?
> > It is related to CustomAppBarButtonLabelToRightStyle:
> > image
> 
> I used [this](https://stackoverflow.com/questions/35937713/xaml-parsing-failed-visual-studio-2013-windows-ui-xaml-markup-xamlparseexcepti) article to create the custom appbar button. I couldn't reproduce the issue however I am removing the key since it isn't affecting the design in any way. Will submit a pull request soon.

It is happening to Windows 10 1809 only. I am guessing this value is not available there.

PR #298, opened 2020-02-21
Did necessary changes pending from previous merge.

soumyamahunt:


JasonStein:
Can you update your branch and pull master? We can discuss here on problems on found in your previous PR.

src/Notepads/Strings/en-US/Resources.resw
```diff
@@ -429,10 +429,6 @@
     <value>selected</value>
     <comment>TextEditor: Singular form for the selected word count indicator.</comment>
   </data>
-  <data name="App_DragAndDrop_UIOverride_Caption_MoveTabHere" xml:space="preserve">
```
JasonStein: Why you remove?

src/Notepads/Controls/TextEditor/TextEditorCore.cs
```diff
@@ -267,15 +267,28 @@ public void SetScrollViewerPosition(double horizontalOffset, double verticalOffs
...
             {
                 _contentLinesCache = (_content + RichEditBoxDefaultLineEnding).Split(RichEditBoxDefaultLineEnding);
                 _isLineCachePendingUpdate = false;
-            }
+            }*/
 
             GetTextSelectionPosition(out var start, out var end);
 
-            lineIndex = 1;
+            lineIndex = (_content + RichEditBoxDefaultLineEnding).Substring(0, start).Length
```
JasonStein: I still think the older way is better. Memory is not a concern here since Notepads cannot handle large file at first place and the string replace here is an overhead when there is no text change during user changing selection and cursor position.

src/Notepads/Strings/en-US/Resources.resw
soumyamahunt: I didn't even touch it, I don't know how it got removed. I am adding it back in.

src/Notepads/Controls/TextEditor/TextEditorCore.cs
soumyamahunt: You can keep the old method but it needs some changes to match selected items with VS Code.

JasonStein: Can you keep the original method as it is and put your new logic in a new method "public void GetCurrentLineColumn2(out int lineIndex, out int columnIndex, out int selectedCount)" and let's decide later since there are already too many changes with in this release. I wan to keep everything minimum. We can then revisit it later.

soumyamahunt: Made changes👍

JasonStein:
Found another bug:

image

If you make some selection and exit the app (make sure you have session snapshot feature enabled), next time when it launches, the selection line is messed up. I am also seeing cases that the hight line is always few lines above the current line.

JasonStein:
Also, the go-to crashes if you enter "0". It has to be handled.

soumyamahunt:
> image
> 
> If you make some selection and exit the app (make sure you have session snapshot feature enabled), next time when it launches, the selection line is messed up. I am also seeing cases that the hight line is always few lines above the current line.

I couldn't reproduce it, have you fixed the bug?

JasonStein:
> > image
> > If you make some selection and exit the app (make sure you have session snapshot feature enabled), next time when it launches, the selection line is messed up. I am also seeing cases that the hight line is always few lines above the current line.
> 
> I couldn't reproduce it, have you fixed the bug?

No, let me try fix it myself, nvm.

PR #301, opened 2020-02-26
Regex feature implemented, fixed goto crash

soumyamahunt:
Implemented find and replace with regular expression.
Fixed goto crash on zero input.

### DIFF
--- a/src/Notepads/Controls/GoTo/GoToControl.xaml.cs
+++ b/src/Notepads/Controls/GoTo/GoToControl.xaml.cs
@@ -122,7 +122,7 @@
                 NotificationCenter.Instance.PostNotification(_resourceLoader.GetString("GoTo_NotificationMsg_InputError_InvalidInput"), 1500);
                 args.Cancel = true;
             }
-            else if (line > _maxLine)
+            else if (line > _maxLine || line == 0)
             {
                 NotificationCenter.Instance.PostNotification(_resourceLoader.GetString("GoTo_NotificationMsg_InputError_ExceedInputLimit"), 1500);
                 args.Cancel = true;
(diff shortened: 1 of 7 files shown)


src/Notepads/Controls/GoTo/GoToControl.xaml.cs
```diff
@@ -122,7 +122,7 @@ private void GoToBar_BeforeTextChanging(TextBox sender, TextBoxBeforeTextChangin
                 NotificationCenter.Instance.PostNotification(_resourceLoader.GetString("GoTo_NotificationMsg_InputError_InvalidInput"), 1500);
                 args.Cancel = true;
             }
-            else if (line > _maxLine)
+            else if (line > _maxLine || line == 0)
```
JasonStein: just make it <=0 to make sure we are 100% safe.

JasonStein:
Can you briefly describe how it works (regex)? And what should we expect, plus some examples?

soumyamahunt:
> Can you briefly describe how it works (regex)? And what should we expect, plus some examples?

In a document open the find control and type your regular expression pattern in the find box.
Select the "Use Regular Expression" in the search options.
Optionally you can select or deselect the "Match Case" to whether take in account or ignore the cases in text respectively.
You can optionally replace one by one or all the words that match the Regular Expression pattern.

For example, open the readme file and give regular pattern "github/[a-z]+" and choose the "Use Regular Expression" feature (keep the "Match Case" unselected). This will give words that match that pattern (in this case user profiles). If you type the pattern "github/[A-Z]+" this will give same result if you keep "Match Case" unselected. If you select match case this would give different result. Alternativly patterns like "\b[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}\b" searches for all emails in a document.

JasonStein:
LGTM, please help test as well.

PR #304, opened 2020-02-28
Made changes to update submenuitem flyout.

soumyamahunt:
Regenerate subitem when recents file list changed.

### DIFF
--- a/src/Notepads/NotepadsMainPage.xaml.cs
+++ b/src/Notepads/NotepadsMainPage.xaml.cs
@@ -175,7 +175,7 @@
             MenuCompactOverlayButton.Click += (sender, args) => EnterExitCompactOverlayMode();
             MenuSettingsButton.Click += (sender, args) => RootSplitView.IsPaneOpen = true;
 
-            MainMenuButtonFlyout.Opening += async (sender, o) =>
+            MainMenuButtonFlyout.Opening += (sender, o) =>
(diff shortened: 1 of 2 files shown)


src/Notepads/NotepadsMainPage.xaml.cs
```diff
@@ -175,7 +175,7 @@ private void InitControls()
             MenuCompactOverlayButton.Click += (sender, args) => EnterExitCompactOverlayMode();
             MenuSettingsButton.Click += (sender, args) => RootSplitView.IsPaneOpen = true;
 
-            MainMenuButtonFlyout.Opening += async (sender, o) =>
```
JasonStein: Why remove async here and use void instead?

soumyamahunt: Since the BuildOpenRecentButtonSubItems() isn't being called inside the opening event handler, there was no reason to keep async there. The reason for removing BuildOpenRecentButtonSubItems() method from there was the flyout wasn't being updated while it was opened (in the next click it was updated). So the rebuilding has to be done before flyout being opened. 
For the first time when app is launched Initcontrols will call BuildOpenRecentButtonSubItems() to populate the recent files values. Then for subsequent file openings, OpenFile method itself will call BuildOpenRecentButtonSubItems().

PR #307, opened 2020-03-01
Recentlist feature

soumyamahunt:
Modified to only show MenuOpenRecentlyUsedFileButton if it contains at least one file path.

### DIFF
--- a/src/Notepads/NotepadsMainPage.xaml.cs
+++ b/src/Notepads/NotepadsMainPage.xaml.cs
@@ -269,21 +269,6 @@
                 }
             }
 
-            openRecentSubItem?.Items?.Add(new MenuFlyoutSeparator());
-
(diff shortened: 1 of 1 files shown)


soumyamahunt:
I think it would be a good idea to show only name of the recent files in the flyout since user can see the complete path from tooltip.

JasonStein:
Firstly, show full path is somewhat a standard across other text editors for recent list feature. Secondly I often edit config files with same name but in different folder. So full path makes more sense in this case. But if more and more users asking for showing filename only, I will add it to the settings to let them choose.

soumyamahunt:
> Secondly I often edit config files with same name but in different folder. So full path makes more sense in this case.

I think in that specific case tooltip will be helpful to show the full path.

PR #316, opened 2020-03-03
Some websearch fixes.

soumyamahunt:
Fixed websearch initiating when no text selected.
When valid url selected web search opens the url instead of searching the text.


JasonStein:
I think it is ok for user to search for empty string, maybe they just want to open the search engine web page?

soumyamahunt:
> I think it is ok for user to search for empty string, maybe they just want to open the search engine web page?

I should remove the empty string check then??

JasonStein:
Also, you need to check for the length of the URI, it will throw if too long:

Invalid URI: The Uri string is too long.

Refernce: https://stackoverflow.com/questions/38440631/httpclient-the-uri-string-is-too-long

You can do substring here to just get first X number of characters just to be sure.

PR #317, opened 2020-03-03
Fixed AppBarButton, superscript toggle on "Ctrl+Shift+="

soumyamahunt:


soumyamahunt:
This is what was happening on "Ctrl+Shift+="

PR #318, opened 2020-03-03
Updated zoom slider style

soumyamahunt:


JasonStein:
Why do we need a new style here? Can you explain a little bit?

soumyamahunt:
> Why do we need a new style here? Can you explain a little bit?

The new style looks like the zoom slider from word, wordpad. It makes more sense and also looks aesthetically pleasing.

JasonStein:
Ok, first, thanks for doing this and it looks not bad to me. There are two concerns I have:

1. I do not like adding more and more custom resources since it might be broken or look wrong on different versions of the OS. All existing custom resources I have added so far are all tint color related or similar reason. I did not actually change their look in anyway because I wanted to keep it the same as the system controls are look like. Everything we do differently will result or cause issues on newer version of the system.

2. If the control changes or have new features added in new Windows SDK, we have to update the resource as well. And that requires human resource and could be troublesome. 

I would suggest let's keep using default control as much as possible and make it easier for us in terms of time and also overall consistency.

soumyamahunt:
> Ok, first, thanks for doing this and it looks not bad to me. There are two concerns I have:
> 
> 1. I do not like adding more and more custom resources since it might be broken or look wrong on different versions of the OS. All existing custom resources I have added so far are all tint color related or similar reason. I did not actually change their look in anyway because I wanted to keep it the same as the system controls. Everything we do differently will result or cause issues on newer or older version of the system (Like the AppBar height issue we faced earlier).
> 2. If the control changes or have new features added in new Windows SDK, we have to update the resource as well. And that requires human resource and could be troublesome.
> 
> I would suggest let's keep using default control as much as possible and make it easier for us in terms of time and also overall consistency.

I have only removed the tint and modified the thumb design, this shouldn't cause any issue in older systems or in future.

JasonStein:
> > Ok, first, thanks for doing this and it looks not bad to me. There are two concerns I have:
> > 
> > 1. I do not like adding more and more custom resources since it might be broken or look wrong on different versions of the OS. All existing custom resources I have added so far are all tint color related or similar reason. I did not actually change their look in anyway because I wanted to keep it the same as the system controls. Everything we do differently will result or cause issues on newer or older version of the system (Like the AppBar height issue we faced earlier).
> > 2. If the control changes or have new features added in new Windows SDK, we have to update the resource as well. And that requires human resource and could be troublesome.
> > 
> > I would suggest let's keep using default control as much as possible and make it easier for us in terms of time and also overall consistency.
> 
> I have only removed the tint and modified the thumb design, this shouldn't cause any issue in older systems or in future.

I am planning to use WinUI at certain point so I wan to keep resource changes as minimum as possible. We should use default design as much as possible for consistency as well unless it is not functional or needed by app logic/ux. Just my 2 cents here.

soumyamahunt:
> I am planning to use WinUI at certain point so I wan to keep resource changes as minimum as possible. We should use default design as much as possible for consistency as well unless it is not functional or needed by app logic/ux. Just my 2 cents here.

This should work with WinUI, here is a comparison of the changes:
zoom-slider-comparison
Rest is up to you, if you don't think this isn't necessary for app ux then let's not do it. Otherwise you can keep this open until WinUI is implemented.

JasonStein:
Let's wait for WinUI to become solid before merging. I don't see it drastically better than what we have right now.

Issue #236, opened 2019-11-04
[Feature request] Hitting Ctrl+F should auto-select everything in the search box


alejandro5042:
**Is your feature request related to a problem? Please describe.**
Super small feature that would significantly improve search UX.

**Describe the solution you'd like**
By hitting Ctrl+F, you want to initiate a new search... but you may have an existing search text in there. So by selecting everything on Ctrl+F, you give the user the option to clear it on the next keystroke, or modify it by pressing an arrow key and then typing.

**Describe alternatives you've considered**
None.

**Additional context**
N/A

entdark:
Yeah, facing the same issue all the time. Very uncomfortable to select all, delete then type again, and sometimes you forget to delete the previous search and it concatenates both.
When hitting Ctrl+F you expect to type a new search query right afterward.

JasonStein:
Fixed and will be available in Notepads v1.1.1.0

alejandro5042:
Awesome! It works great. Thank you 👍

PR #319, opened 2020-03-03
Now text is selected when findbar comes into focus.

soumyamahunt:


soumyamahunt:
@JasonStein [this](https://github.com/JasonStein/Notepads/issues/236) issue can be closed now.

PR #321, opened 2020-03-03
Selected count now matches exactly to that of VS code in all scenarios.

soumyamahunt:


JasonStein:
We do not have to match VSCode 100% since WinRT richest behaves differently and it actually select the end string as well. In other words, the current selection count math is correct based on what we observed.

soumyamahunt:
> We do not have to match VSCode 100% since WinRT rich edit behaves differently and it actually select the end char as well. In other words, the current selection count math is correct based on what we observed.

The weird behavior I have seen is the selected count is increased by one when the line ending character of the first selected line is selected. But for subsequent lines the line ending character increases selection count by two. But for VS Code line ending character is always counted as two.

Another thing is with RichEditBox adds line ending character at the end of the document which is not the case for VS Code. I don't count it in this request but I can change the code to count the last line ending character.

soumyamahunt:
> We do not have to match VSCode 100% since WinRT rich edit behaves differently and it actually select the end char as well. In other words, the current selection count math is correct based on what we observed.

Example of the issue with selected count:
selection-example-1
selection-example-2

JasonStein:
Will this change affect DuplicateText or Indentation change?

soumyamahunt:
> Will this change affect DuplicateText or Indentation change?

Since selectedcount isn't used in DuplicateText or IndentationChange, it won't affect them.

JasonStein:
@soumyamahunt  I just realized that why I don't want to fix it. VS selected count will change based on line ending. This is only going to match the case where line ending is CRLF. If you change line ending to LF or CR, it matches what we have in Notepads.

soumyamahunt:
> @soumyamahunt I just realized that why I don't want to fix it. VS selected count will change based on line ending. This is only going to match the case where line ending is CRLF. If you change line ending to LF or CR, it matches what we have in Notepads.

In the previous case selected count didn't match even for LF/CR line ending. I have made changes to match selected count for both LF and CRLF line ending. In VS Code CR line ending is detected as CRLF so I made selected count in CR line ending to match to that of LF line ending.

PR #322, opened 2020-03-03
Log entry feature

soumyamahunt:
Implemented system notepad style log entry.
Provided option in settings to disable this feature.


JasonStein:
I still do not think we should add this feature at all. I am leaving it open for now before my final decision.

soumyamahunt:
> I still do not think we should add this feature at all. I am leaving it open for now before my final decision.

If more people from community request this, you can add this feature:)

JasonStein:
> > I still do not think we should add this feature at all. I am leaving it open for now before my final decision.
> 
> If more people from community request this, you can add this feature:)

Yes, I will keep this PR open.

shanselman:
I just got this email after promoting Notepads in an newsletter:

"Scott, downloaded NotePads.  Too bad they did not include the .LOG code in the new app. If you add .LOG on the first line of a txt file if will embed the date and time that you open it each time.  
.LOG
8:10 AM 7/23/2020
Please tell the Notepads team to include the code the original team use a zillion years ago.
I use this feature on many documents each day."

Yes it's obscure, but it's harmless and fun and for the few people who are obsessed, why not add it?

src/Notepads/Controls/TextEditor/TextEditorCore.DateTime.cs
```diff
@@ -6,11 +6,28 @@
...
         {
             var dateStr = DateTime.Now.ToString(CultureInfo.CurrentCulture);
             Document.Selection.SetText(TextSetOptions.None, dateStr);
             Document.Selection.StartPosition = Document.Selection.EndPosition;
         }
+
+        public void TryInsertNewLogEntry()
+        {
+            var docText = GetText();
+            if (!docText.StartsWith(".LOG") || _hasAddedLogEntry) return;
```
JasonStein, line 21: check _hasAddedLogEntry first in the first line so you can skip getting text in some cases.

soumyamahunt: Done.

src/Notepads/Controls/TextEditor/TextEditorCore.DateTime.cs
```diff
@@ -6,11 +6,28 @@
...
+        private bool _hasAddedLogEntry = false;
+
         private void InsertDateTimeString()
         {
             var dateStr = DateTime.Now.ToString(CultureInfo.CurrentCulture);
             Document.Selection.SetText(TextSetOptions.None, dateStr);
             Document.Selection.StartPosition = Document.Selection.EndPosition;
         }
+
+        public void TryInsertNewLogEntry()
```
JasonStein, line 18: Add a summary to this method and paste some links describing this feature since only few people knows about it.

soumyamahunt: Done.

src/Notepads/Controls/TextEditor/TextEditorCore.DateTime.cs
```diff
@@ -6,11 +6,36 @@
...
+        /// Adds "Windows Notepad" style header with current date and time, 
+        /// if a text document contains ".LOG" at the very beginning of file.
+        /// User can then add log entry from the very next line.
+        /// </para>
+        /// <see cref="https://support.microsoft.com/help/260563/how-to-use-notepad-to-create-a-log-file"/>
+        /// </summary>
+        public void TryInsertNewLogEntry()
+        {
+            var docText = _hasAddedLogEntry ? string.Empty : GetText();
+            if (string.IsNullOrEmpty(docText) || !docText.StartsWith(".LOG")) return;
```
JasonStein, line 29: docText.StartsWith(".LOG")
What about a file looks like this:
-------
.LOG123456
7890
-------
Is this a valid .LOG file or no? Should we use .StartsWith(".LOG\r") here?

shanselman:
Yes, confirmed with standard notepad.exe that it adds the date if
"StartsWith()"

src/Notepads/Controls/TextEditor/TextEditorCore.DateTime.cs
soumyamahunt: > ## .LOG123456
> 7890
> Is this a valid .LOG file or no?

Yes, this is treated as valid .LOG file in default Notepad.

shanselman:
:shipit:

PR #364, opened 2020-03-19
Fixed drag-drop not saving file without filesavepicker(User has to manually give permission to access "File System" in settings).

soumyamahunt:


soumyamahunt:
@JasonStein should we prompt user to open settings dialog on startup??

JasonStein:
Let's not use "broadFileSystemAccess" for now.
1. It will confuse user and hard to get app reviewed/passed in store.
2. Even with broadFileSystemAccess, there are still places that you have no access to.

soumyamahunt:
> Let's not use "broadFileSystemAccess" for now.
> 
> 1. It will confuse user and hard to get app reviewed/passed in store.
> 2. Even with broadFileSystemAccess, there are still places that you have no access to.

Three scenarios I think "broadFileSystemAccess" will be helpful are:

1. In this case to save file without filesavepicker.
2. Render images in MD view in local folders.
3. Optionally create file when passed path through commandline.

In future it would be necessary, so even if you don't merge it now, I would suggest to keep it open.

JasonStein:
> > Let's not use "broadFileSystemAccess" for now.
> > 
> > 1. It will confuse user and hard to get app reviewed/passed in store.
> > 2. Even with broadFileSystemAccess, there are still places that you have no access to.
> 
> Three scenarios I think "broadFileSystemAccess" will be helpful are:
> 
> 1. In this case to save file without filesavepicker.
> 2. Render images in MD view in local folders.
> 3. Optionally create file when passed path through commandline.
> 
> In future it would be necessary, so even if you don't merge it now, I would suggest to keep it open.

Yes, I have tried it once before and give up because of the mess it brings. I used to have this flag enabled during preview but users are afraid to use since it asks for full disk access. Considering the target audience of notepads are normal people, who knows little about compute programming. They probably not going to give permission or will delete the app right away.

PR #366, opened 2020-03-19
Hot-fix: Fixed tab not being inserted in new empty line.

soumyamahunt:


JasonStein:
This is actually a bug of GetLineColumnSelection, we should fix there.

soumyamahunt:
> This is actually a bug of GetLineColumnSelection, we should fix there.

Changing GetLineColumnSelection, breaks multilineindentation. For now, this fix should make everything work as expected until I change the indentation methods.

JasonStein:
Found another issue with duplicate text:

image

Ctrl+d:

image

PR #375, opened 2020-03-21
Markdown preview now detects local files better

soumyamahunt:
Implemented image preview for local image file in markdown.
Implemented opening local files by markdown links.


soumyamahunt:
@JasonStein this implements "broadFileSystemAccess" access too. So no need to merge it right away.

Issue #300, opened 2020-02-22
[Bug] Notepads doesn't open through git bash


fuadnafiz98:
#### Input 

```sh
$ notepads
```

#### Output

```sh
 bash: /c/Users/USER/AppData/Local/Microsoft/WindowsApps/notepads: Permission denied
```

JasonStein:
Is it official git command line? Can you try run under admin privilege?

soumyamahunt:
> Is it official git command line? Can you try run under admin privilege?

git-cmd works fine, the issue is with git-bash.

soumyamahunt:
> ### If I try to open notepads with git-bash command line, it shows the following error
> #### Input
> ```shell
> $ notepads
> ```
> 
> #### Output
> ```shell
>  bash: /c/Users/USER/AppData/Local/Microsoft/WindowsApps/notepads: Permission denied
> ```

Type `powershell.exe -c notepads` instead of `notepads `to fix it.

PR #379, opened 2020-03-22
Support opening files inside wsl environments

soumyamahunt:


JasonStein:
Is it WSL2?

soumyamahunt:
> Is it WSL2?

Both should work, in wsl environments run

`powershell.exe /c notepads your/linux/path`

Use linux default path separater character if you are giving path without quotes.

JasonStein:
should use  powershell.exe -c notepads instead /c to not block bash.

soumyamahunt:
> should use powershell.exe -c notepads instead /c to not block bash.

[This issue](https://github.com/JasonStein/Notepads/issues/300) can be fixed using above command. Should mark it fixed.

Issue #237, opened 2019-11-04
[Bug] Shift-F3 should go to the previous search hit


alejandro5042:
**Describe the bug**
Right now Shift+F3 does some casing transformations. This is unusual and non-standard.

**To Reproduce**
Search some text and hit Shift+F3.

**Expected behavior**
Go to the previous hit.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Windows 10

**Additional context**
You can move the casing feature to another F-key. That's OK with me.

JasonStein:
This has been implemented. Closing the thread for now.

alejandro5042:
Awesome! thank you!

Issue #139, opened 2019-08-13
[Feature request] Get to previous result when searching


TailyFair:
**Is your feature request related to a problem? Please describe.**
There is no way to get back to previous search result.

**Describe the solution you'd like**
While you can go forward to next search result using F3, you can't go back to previous result.

**Describe alternatives you've considered**
Maybe it would be possible to add additional shortcut that goes to previous result.

Also it would be convenient to add UI arrows in Search bar to navigate forwards and backwards.

JasonStein:
I implemented the current Find and Replace solution in only a few hours a while back and have not pay much attention to it since then. I will come back later for this but yes, improvements are coming.

JasonStein:
This has been implemented. Closing the thread for now.

Issue #120, opened 2019-08-09
[Feature request] Search Enhancement


Threonine-Animist:
1.It might be better if Notepads can search instantly when some words is typed in searchbox.
Making the matched highlighted in the scrollbar also sounds great. Just like what Chrome does.
2.Please make the availability of F3 function all the time. In current version F3 is available only when the searchbox is open.

JasonStein:
For option 1, I have a plan but not sure if it is doable by design, I will try my best anyway.
For option 2, let me think about it. It is a valid suggestion tho.

United600:
Not to hijack this issue but the ability to switch input fields with the Shift key would rock

JasonStein:
Enhancements have been made to Find and Replace. Closing this thread for now.

PR #386, opened 2020-03-24
Some enhancement to find-replace functionality

soumyamahunt:
1. Added reverse search with Shift+F3 as requested in issue #237 and #139.
2. Added reverse search button and modifies find-replace control design as requested in issue #139.
0F6OLYRd7p
ApplicationFrameHost_QvsVXFDthi

3. Modified some colors.
4. Added remembering search option to do search even after control dismissed requested in second part of #120.
5. Added pressing enter in Replacebar to replace.


src/Notepads/Controls/FindAndReplace/FindAndReplaceControl.xaml.cs
```diff
@@ -68,9 +68,12 @@ private void SetSelectionHighlightColor()
                 Application.Current.Resources["SystemControlForegroundAccentBrush"] as SolidColorBrush;
         }
 
-        public void Focus()
+        public void Focus(FindAndReplaceMode mode)
         {
-            FindBar.Focus(FocusState.Programmatic);
+            if(mode==FindAndReplaceMode.FindOnly)
```
JasonStein: mode == FindAndReplaceMode.FindOnly

src/Notepads/Controls/FindAndReplace/FindAndReplaceEventArgs.cs
```diff
@@ -9,16 +9,23 @@ public enum FindAndReplaceMode
         ReplaceAll
     }
 
+    public enum Direction
```
JasonStein: Rename it to SearchDirection and use "Previous" or "Next" here. The reason we should not use down or up is because we are assuming text flow is always top to down which might not be true for some languages. I could not think of any but there is a potential, similar like Right to Left languages. Previous or Next seems more accurate here.

JasonStein:
The reverse search works for normal words but not working for single character. Try search for a single character and hit previous. It won't work.

soumyamahunt:
The updated ui:
0F6OLYRd7p
ApplicationFrameHost_QvsVXFDthi

JasonStein:
> The updated ui:
> 0F6OLYRd7p
> ApplicationFrameHost_QvsVXFDthi

Nice, can you shrimp the size of the left button, as well as the icon size:
image

src/Notepads/Controls/FindAndReplace/FindAndReplaceControl.xaml
```diff
@@ -4,81 +4,156 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:querybox="using:Notepads.Controls.FindAndReplace"
     mc:Ignorable="d"
     d:DesignHeight="72"
     d:DesignWidth="250">
 
-    <Grid x:Name="RootGrid">
+    <Grid x:Name="RootGrid" KeyDown="RootGrid_KeyDown">
```
JasonStein: Name it "FindAndReplaceRootGrid" and change event name to "FindAndReplaceRootGrid_OnKeyDown"

src/Notepads/Controls/FindAndReplace/FindAndReplaceControl.xaml
```diff
@@ -4,81 +4,156 @@
...
+                    IsTabStop="False"
+                    Width="30"
+                    FontSize="10"
+                    Style="{StaticResource DismissButtonStyle}"
+                    Content="&#xE00F;"
+                    FontFamily="Segoe MDL2 Assets"
+                    Click="ToggleReplaceModeButton_OnClick">
+            </Button>
+        </Grid>
+
```
JasonStein: Remove empty line.

src/Notepads/Controls/FindAndReplace/FindAndReplaceControl.xaml
```diff
@@ -4,81 +4,156 @@
...
             <RowDefinition Height="Auto" />
         </Grid.RowDefinitions>
 
-        <Grid Grid.Row="0"  Padding="1, 1, 1, 1" Height="36" x:Name="FindAndReplaceRootGrid">
+        <Grid Grid.Row="0" Grid.Column="0" Padding="1, 1, 1, 1" Height="36" x:Name="ToggleReplaceModeButtonGrid">
+            <Button x:Name="ToggleReplaceModeButton"
+                    x:Uid="FindAndReplace_ToggleReplaceModeButton"
+                    Grid.Column="0"
+                    IsTabStop="False"
+                    Width="30"
```
JasonStein: Width="25"
Padding="0"
FontSize="12"

src/Notepads/Controls/FindAndReplace/FindAndReplaceControl.xaml.cs
```diff
@@ -162,5 +223,27 @@ private void OptionButtonFlyoutItem_OnClick(object sender, RoutedEventArgs e)
...
             UseRegexToggle.IsEnabled = !MatchWholeWordToggle.IsChecked;
         }
+
+        private void RootGrid_KeyDown(object sender, KeyRoutedEventArgs e)
+        {
+            var ctrl = Window.Current.CoreWindow.GetKeyState(VirtualKey.Control);
+            var alt = Window.Current.CoreWindow.GetKeyState(VirtualKey.Menu);
+            var shift = Window.Current.CoreWindow.GetKeyState(VirtualKey.Shift);
+
+            if (!(!ctrl.HasFlag(CoreVirtualKeyStates.Down) && !alt.HasFlag(CoreVirtualKeyStates.Down) && !shift.HasFlag(CoreVirtualKeyStates.Down) && e.Key == VirtualKey.Escape) &&
```
JasonStein: Use KeyboardCommandHandler + KeyboardShortcut here for better readability.

src/Notepads/Controls/GoTo/GoToControl.xaml.cs
```diff
@@ -128,5 +131,18 @@ private void GoToBar_BeforeTextChanging(TextBox sender, TextBoxBeforeTextChangin
                 args.Cancel = true;
             }
         }
+
+        private void RootGrid_KeyDown(object sender, KeyRoutedEventArgs e)
```
JasonStein: Use KeyboardCommandHandler + KeyboardShortcut here for better readability.

src/Notepads/Resource/TransparentTextBoxStyle.xaml
```diff
@@ -6,12 +6,12 @@
     <ResourceDictionary.ThemeDictionaries>
         <ResourceDictionary x:Key="Light">
             <SolidColorBrush x:Key="CustomTextBoxBackground"
-                            Color="#E0E0E0"
+                            Color="#C0C0C0"
```
JasonStein: Let's not change this for now. The color difference between mouse over and selected is visually interactive. Also the grey color does not look good in Light theme.

src/Notepads/Resource/TransparentTextBoxStyle.xaml
```diff
@@ -6,12 +6,12 @@
...
                             Opacity="0.7"/>
         </ResourceDictionary>
         <ResourceDictionary x:Key="Dark">
             <SolidColorBrush x:Key="CustomTextBoxBackground"
-                             Color="#1E1E1E"
+                             Color="#2E2E2E"
```
JasonStein: Same here, revert for now.

src/Notepads/Resource/TransparentTextBoxStyle.xaml
```diff
@@ -78,12 +78,12 @@
                                 </VisualState>
                                 <VisualState x:Name="PointerOver">
                                     <Storyboard>
-                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="BorderElement" Storyboard.TargetProperty="BorderBrush">
+                                        <!--<ObjectAnimationUsingKeyFrames Storyboard.TargetName="BorderElement" Storyboard.TargetProperty="BorderBrush">
```
JasonStein: I would suggest to revert this file.

src/Notepads/Strings/en-US/Resources.resw
```diff
@@ -573,4 +573,12 @@
     <value>Encoding cannot be determined</value>
     <comment>TextEditor: Notification message when file's encoding cannot be determined.</comment>
   </data>
+  <data name="FindAndReplace_SearchBackwardButton.ToolTipService.ToolTip" xml:space="preserve">
```
JasonStein: Can you add these two newly added entries to all locales? Just use English for all.

src/Notepads/Resource/TransparentTextBoxStyle.xaml
soumyamahunt: I wanted to make the options button background same as the textbox background. The only way I could achieve it was by having same color for normal, mouse over and selected case. I have reverted the changes but I couldn't find a better way to sync the colors. Do you know any way it can be done??

src/Notepads/Controls/GoTo/GoToControl.xaml.cs
```diff
@@ -15,12 +17,16 @@ public sealed partial class GoToControl : UserControl
...
 
+        public event EventHandler<KeyRoutedEventArgs> OnGoToControlKeyDown;
+
         private int _currentLine;
 
         private int _maxLine;
 
         private readonly ResourceLoader _resourceLoader = ResourceLoader.GetForCurrentView();
 
+        private readonly KeyboardShortcut<bool> _nativeKeyboardShortcut = new KeyboardShortcut<bool>(false, false, false, VirtualKey.Escape, null);
```
JasonStein: I don't get this, can you explain?

src/Notepads/Controls/TextEditor/TextEditor.xaml.cs
```diff
@@ -857,37 +908,50 @@ public void HideFindAndReplaceControl()
...
+        {
+            if (string.IsNullOrEmpty(findAndReplaceEventArgs.SearchText)) return;
+
             bool found = false;
-            string regexError = null;
+            bool regexError = false;
 
-            switch (e.FindAndReplaceMode)
+            if (FindAndReplacePlaceholder?.Visibility == Visibility.Visible)
+                _searchCache = new SearchCache(findAndReplaceEventArgs.SearchText, findAndReplaceEventArgs.MatchCase, findAndReplaceEventArgs.MatchWholeWord, findAndReplaceEventArgs.UseRegex);
```
JasonStein: Is the _searchCache intended to carry current search settings to all newly added editors?  If yes, we might need to not do it for now. Reason: there is no UI to show what search options user has been made. I am think about the design here, will let you know once I come up with one.

soumyamahunt: > Is the _searchCache intended to carry current search settings to all newly added editors?  If yes, we might need to not do it for now. Reason: there is no UI to show what search options user has been made. I am think about the design here, will let you know once I come up with one.

No every editor will have thier separate search results in `_searchcache` and for new editors it would have empty search string with default options.

src/Notepads/Controls/GoTo/GoToControl.xaml.cs
soumyamahunt: > I don't get this, can you explain?

The `_native KeyboardShortcuts`/`_nativekeyboardShortcut` variables store all the keyboard shortcut of FindReplace/GoTo control respectively. In the event of key down, the keyboard shortcut is compared with it and if the provided shortcut is not implemented then the shortcut is passed to the editor which then handles it.
The reason I didn't implement KeyboardHandler is that it will only work when the control is on focus. Hence it is better to have shortcuts handled in xaml and editor shortcuts passed via the keydown event.

JasonStein: Sorry, I mean this Esc thingy.

soumyamahunt: You are right, I added this to handle the Esc key action in the control itself. But since it is already handled in editor we can remove it.

JasonStein:
Ready to merge?

soumyamahunt:
> Ready to merge?

Let me have a last look.

JasonStein:
> > Ready to merge?
> 
> Let me have a last look.

Can you also slightly increase the width of the find and replace bar? Since now it is more crowed. I would suggest 320 (what we have right now) -> 360
https://github.com/JasonStein/Notepads/blob/a77ff0040b5baf1c668d3799f47dcde9ac709e49/src/Notepads/Controls/TextEditor/TextEditor.xaml#L119

src/Notepads/Controls/TextEditor/TextEditorCore.cs
```diff
@@ -935,6 +935,35 @@ private static int LastIndexOfWholeWord(string target, int startIndex, string va
...
+        public string GetSearchString()
+        {
+            var searchString = Document.Selection.Text.Trim();
+
+            if (searchString.Contains(RichEditBoxDefaultLineEnding.ToString())) return string.Empty;
+
+            if (string.IsNullOrEmpty(searchString) && Document.Selection.StartPosition < _content.Length)
+            {
+                var startIndex = Document.Selection.StartPosition;
+                var endIndex = Document.Selection.StartPosition;
```
JasonStein: Document.Selection.EndPosition

soumyamahunt: This case is only triggered either when no text is selected(In that case it won't matter) or when only spaces, tabs, line ending characters selected in that case it would produce unnecessary results. I would suggest to keep it same.

soumyamahunt:
> Ready to merge?

I am satisfied with the changes, let me know if you want to change anything.

JasonStein:
> > Ready to merge?
> 
> I am satisfied with the changes, let me know if you want to change anything.

Instead of add "bool passSearchString", can you do "string searchString" instead and pass the real search string if any? This will make the API looks better and more readable. passSearchString is a little bit confusing since searchString is out of context here.

soumyamahunt:
> > > Ready to merge?
> > 
> > 
> > I am satisfied with the changes, let me know if you want to change anything.
> 
> Instead of add "bool passSearchString", can you do "string searchString" instead and pass the real search string if any? This will make the API looks better and more readable. passSearchString is a little bit confusing since searchString is out of context here.

While `searchString ` is used pass the search string to findreplace, `passSearchString` handles whether or not pass the search string or not. For example, when user clicks on the toggle replace mode button or presses find-replace keyboard shortcut from inside of findreplace control the search string won't be passed. I think this is essential for optimum user experience.

JasonStein:
> > > > Ready to merge?
> > > 
> > > 
> > > I am satisfied with the changes, let me know if you want to change anything.
> > 
> > 
> > Instead of add "bool passSearchString", can you do "string searchString" instead and pass the real search string if any? This will make the API looks better and more readable. passSearchString is a little bit confusing since searchString is out of context here.
> 
> While `searchString ` is used pass the search string to findreplace, `passSearchString` handles whether or not pass the search string or not. For example, when user clicks on the toggle replace mode button or presses find-replace keyboard shortcut from inside of findreplace control the search string won't be passed. I think this is essential for optimum user experience.

Yes, I understand why you need it. In this case, you should handle this logic within the find and replace control instead of letting external caller to decide. You can store the "view" state in find and replace to make it context aware. You can reset the flag whenever find and replace is been hidden.

src/Notepads/Controls/FindAndReplace/FindAndReplaceControl.xaml.cs
```diff
@@ -35,6 +35,8 @@ public sealed partial class FindAndReplaceControl : UserControl
         //This variable is used to remove flicker in text selection
         private bool _enterPressed = false;
 
+        private bool _updateSearchString = true;
```
JasonStein: _shouldUpdateSearchString

src/Notepads/Controls/FindAndReplace/FindAndReplaceControl.xaml
```diff
@@ -4,143 +4,202 @@
...
                         Width="36"
-                        Grid.Column="2" 
+                        Grid.Column="1" 
                         FontSize="16"
                         Style="{StaticResource DismissButtonStyle}"
                         Content="&#xE7FD;" 
                         FontFamily="Segoe MDL2 Assets"
                         Click="ReplaceAllButton_OnClick">
+                    <Button.KeyboardAccelerators>
+                        <KeyboardAccelerator Modifiers="Menu" Key="A"/>
```
JasonStein: Alt+A is reserved by Window clipping tool. Ctrl+shift+enter is good enough here.

src/Notepads/Strings/cs-CZ/Resources.resw
```diff
@@ -206,18 +206,18 @@
     <comment>FindAndReplace: Notification message when target not found.</comment>
   </data>
   <data name="FindAndReplace_ReplaceAllButton.ToolTipService.ToolTip" xml:space="preserve">
-    <value>Nahradit vše</value>
+    <value>Nahradit vše (Alt+A)</value>
```
JasonStein: Change all these to "Ctrl+Shift+Enter"

src/Notepads/Controls/FindAndReplace/FindAndReplaceControl.xaml.cs
```diff
@@ -13,6 +16,27 @@ public sealed partial class FindAndReplaceControl : UserControl
...
 
+        public event EventHandler<bool> OnToggleReplaceModeButtonClicked;
+
+        public event EventHandler<KeyRoutedEventArgs> OnFindReplaceControlKeyDown;
+
+        private readonly List<KeyboardShortcut<bool>> _nativeKeyboardShortcuts = new List<KeyboardShortcut<bool>>
+        {
+                new KeyboardShortcut<bool>(VirtualKey.F3, null),
+                new KeyboardShortcut<bool>(false, false, true, VirtualKey.F3, null),
+                new KeyboardShortcut<bool>(false, true, false, VirtualKey.A, null),
```
JasonStein: Remove Ctrl+A here as well.

src/Notepads/Controls/FindAndReplace/FindAndReplaceControl.xaml
soumyamahunt: By window clipping tool do you mean snipping tool?? I can't find when it uses this shortcut. I used "Alt+A" cause notepad uses it. "Ctrl+Shift+Enter" was just to match funtionality with VS Code. Wouldn't users expect "Alt+A" to do this by default??

JasonStein: Oh, I see. Actually it was reserved by another program that I am using. Nvm.

soumyamahunt: Should I change it??

JasonStein: Let's just remove alt+A since notepad is too old. Let's follow the trend of VS Code here.

soumyamahunt: Just noticed that notepad++ and vs studio use "Alt+A" as well, may be other text editors do this too.

JasonStein: Let's have Ctrl+Shift+Enter for now and see how user response. Most of the people don't use shortcut at all. We should be fine.

src/Notepads/Controls/TextEditor/TextEditor.xaml.cs
```diff
@@ -36,6 +36,25 @@ public enum FileModificationState
         RenamedMovedOrDeleted
     }
 
+    public class SearchCache
```
JasonStein: Move to single file and put it under FindAnReplace

src/Notepads/Controls/FindAndReplace/FindAndReplaceControl.xaml
soumyamahunt: I have removed Alt+A, so it is ready to be merged.

src/Notepads/Controls/TextEditor/TextEditor.xaml.cs
JasonStein: I will merge after this change. Thanks for the hard work!

JasonStein:
Found a bug:
image

Hit replace all using "Ctrl+Shift+Enter" will result in "Not Found"

soumyamahunt:
> Found a bug:
> image
> 
> Hit replace all using "Ctrl+Shift+Enter" will result in "Not Found"

The actual shortcut is Ctrl+Alt+Enter, I have pushed a fix.

PR #391, opened 2020-03-27
Add find-on-selection functionality

krlvm:
If when the user starts the search there is selected text, then the search text will be this selection.
This is implemented in Notepad, Chrome, etc.
Demo (I'm using Ctrl+F):
selectfind


JasonStein:
Another contributor is working on improving find and replace at this moment, you might need to update your PR later. Just let you know. By the way, we need to set a maximum length on this feature. I would not suggest to search on string that is greater than 100 in length.

soumyamahunt:
> Another contributor is working on improving find and replace at this moment, you might need to update your PR later. Just let you know. By the way, we need to set a maximum length on this feature. I would not suggest to search on string that is greater than 100 in length.

I have implemented slightly modified version in my PR.

PR #395, opened 2020-03-28
RTL supports

JasonStein:
Added RTL button in context menu. Properly fixed LTR<->RTL display/rendering issue.

### DIFF
--- a/src/Notepads/Controls/TextEditor/TextEditorCore.cs
+++ b/src/Notepads/Controls/TextEditor/TextEditorCore.cs
@@ -703,8 +703,8 @@
 
             FlowDirection = direction;
             TextReadingOrder = TextReadingOrder.UseFlowDirection;
-            TextReadingOrder = TextReadingOrder.DetectFromContent;
 
+            UpdateLayout();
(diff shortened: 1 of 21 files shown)


src/Notepads/Controls/TextEditor/TextEditorCore.cs
```diff
@@ -690,6 +690,20 @@ public bool FindPrevAndSelect(string searchText, bool matchCase, bool matchWhole
...
+            if (string.IsNullOrEmpty(_content))
+            {
+                // If content is empty, switching text flow direction might not work
+                // Let's not do anything here
+                return;
+            }
+
+            FlowDirection = direction;
+            TextReadingOrder = TextReadingOrder.UseFlowDirection;
+            TextReadingOrder = TextReadingOrder.DetectFromContent;
```
be5invis: You can remove this now since the BiDi issue is mainly caused by `TextScript`-thingy.
The assignment `TextReadingOrder = TextReadingOrder.DetectFromContent` in the constructor could also be removed.

be5invis: Please remove this following my PR. Otherwise changing direction may not work.

PR #452, opened 2020-04-20
Enabled settings syncing accross all windowss

soumyamahunt:
Settings is now synced for all app instances.

## PR Type

Feature


soumyamahunt:
@JasonStein all the settings except sessionsnapshot are synced, let me know if you want to change anything.

JasonStein:
Thanks for the PR. I was hesitated on implementing any background task for Notepads because background task are not guaranteed to work in all circumstances. It might fail to start and it might be disabled by user. Like I literally disabled background tasks for some of my daily used UWP apps. If that happens, settings will not be able to work at all right?

soumyamahunt:
> Thanks for the PR. I was hesitated on implementing any background task for Notepads because background task are not guaranteed to work in all circumstances. It might fail to start or even disabled in battery saver mode. And it might also be disabled by user manually like I literally disabled background tasks for some of my daily used UWP apps. If that happens, settings will not be able to work at all right? That's also why I did no implement background task for checking file status and instead I do it in main process.

Tried with disabling background apps and turning on battery saver mode, it works as expected.

JasonStein:
> > Thanks for the PR. I was hesitated on implementing any background task for Notepads because background task are not guaranteed to work in all circumstances. It might fail to start or even disabled in battery saver mode. And it might also be disabled by user manually like I literally disabled background tasks for some of my daily used UWP apps. If that happens, settings will not be able to work at all right? That's also why I did no implement background task for checking file status and instead I do it in main process.
> 
> Tried with disabling background apps and turning on battery saver mode, it works as expected.

Is this task launched and getting killed as part of the main process? Or it runs indefinitely?

soumyamahunt:
> Is this task launched and getting killed as part of the main process? Or it runs indefinitely?

The task runs in a separate process, it stays alive as long as the app is running.

JasonStein:
I still this is overkill for a small benefit. We are making it complicated for settings and now we will have another process running. I think we should hold off until we find another use case of it to justify the cost and maintainability out of it. Sorry for saying so because this is a huge amount of work that you did. Let's keep this PR alive so we can reuse them in the future if needed.

soumyamahunt:
> I still think this is overkill for a small benefit. We are making it complicated for settings and now we will have another process running. I think we should hold off until we find another use case of it to justify the cost and maintainability out of it. Sorry for saying so because this is a huge amount of work that you did. Let's keep this PR alive so we can reuse them in the future if needed.

Not being able to access settings in shadow window is a real inconvenience for me, also implementing appservice will also be helpful implementing desktop extension. Do you still want to implement desktop extension or have you decided against that??

JasonStein:
> > I still think this is overkill for a small benefit. We are making it complicated for settings and now we will have another process running. I think we should hold off until we find another use case of it to justify the cost and maintainability out of it. Sorry for saying so because this is a huge amount of work that you did. Let's keep this PR alive so we can reuse them in the future if needed.
> 
> Not being able to access settings in shadow window is a real inconvenience for me, also implementing appservice will also be helpful implementing desktop extension. Do you still want to implement desktop extension or have you decided against that??

I originally wanted to use desktop extension but slowly I found it is not necessary especially when Windows 10X was announced. The main bottleneck we have is something like the performance of the WinRT RichEdit and also the lack of large file support. Other things are not so important. I also wanted to keep Notepads as minimum as possible not just in terms of design but also the amount of code. So it is easier for me to maintain and do drastic changes in the future. 

On the other hands, the majority of the users never changed anything in settings and settings are not so important to their daily usage of the app. As well for multi instances, let me show you the usage stats:

Screen Shot 2020-04-25 at 5 29 55 PM

For the past 90days, only 1% of the usage goes for shadow window (non-first instance). That is the count of window, the real user base can only be lower.

soumyamahunt:
> I originally wanted to use desktop extension but slowly I found it is not necessary especially when Windows 10X was announced. 

I have added desktop extension to allow user to modify system files directly. You can check it out [here](https://github.com/soumyamahunt/Notepads/tree/desktop-extension). I personally use it so I don't have to look at "MS Notepad" when I modify system files like hosts file. Also cases are considered for Windows 10X where desktop extension is not available, so the app won't fail there. I completely understand if you don't want to implement it since most users likely will never use it.

> For the past 90days, only 1% of the usage goes for shadow window (non-first instance). That is the count of window, the real user base can only be lower.

For those of us that use Windows' virtual desktop feature shadow window is a must. Again if you implement this I think this would increase the user base.

JasonStein:
> > I originally wanted to use desktop extension but slowly I found it is not necessary especially when Windows 10X was announced.
> 
> I have added desktop extension to allow user to modify system files directly. You can check it out [here](https://github.com/soumyamahunt/Notepads/tree/desktop-extension). I personally use it so I don't have to look at "MS Notepad" when I modify system files like hosts file. Also cases are considered for Windows 10X where desktop extension is not available, so the app won't fail there. I completely understand if you don't want to implement it since most users likely will never use it.
> 
> > For the past 90days, only 1% of the usage goes for shadow window (non-first instance). That is the count of window, the real user base can only be lower.
> 
> For those of us that use Windows' virtual desktop feature shadow window is a must. Again if you implement this I think this would increase the user base.

Feel free to send out PR and I can take a look. I am hesitate on using background service to sync settings but I do want to leverage desktop extension to enable editing on host files etc or use it to inject Registry so it can add command to right click context menu.

soumyamahunt:
> Feel free to send out PR and I can take a look. I am hesitate on using background service to sync settings but I do want to leverage desktop extension to enable editing on host files etc or use it to inject commands to Registry so we can leverage right click context menu in shell.

I have submitted the pr, the background process is a must for communicating between desktop extension and uwp. Syncing settings is just an added benifit.

JasonStein:
> > Feel free to send out PR and I can take a look. I am hesitate on using background service to sync settings but I do want to leverage desktop extension to enable editing on host files etc or use it to inject commands to Registry so we can leverage right click context menu in shell.
> 
> I have submitted #523 , the background process is a must for communicating between desktop extension and uwp. Syncing settings is just an added benifit.

If that's the case, you can combine them in one PR. Btw, do you know if Windows 10X will support desktop extension?

I will probably not going to spent too much time on this PR until the XBox game bar extension is released since there are some packaging issues that I need to resolve.

soumyamahunt:
> If that's the case, you can combine them in one PR.

Will do that.

> Btw, do you know if Windows 10X will support desktop extension?
 
Probably not, but I have made changes in code so that it checks if desktop extension api is available otherwise it works normally as before

> I will probably not going to spent too much time on this PR until the XBox game bar extension is released since there are some packaging issues that I need to resolve.

Let me know when you decide to merge it.

PR #465, opened 2020-04-24
Fixed SetDraggedOutside limit.

soumyamahunt:
Fixed SetDraggedOutside limit. Dragging any type of set item out side the app will launch a new instance and open that set in the instance.

## PR Type

Feature

### DIFF
--- a/src/Notepads/Core/NotepadsCore.cs
+++ b/src/Notepads/Core/NotepadsCore.cs
@@ -24,6 +24,7 @@
     using Windows.UI.Xaml.Input;
     using Windows.UI.Xaml.Media;
     using Microsoft.AppCenter.Analytics;
+    using Notepads.Core.SessionDataModels;
 
     public class NotepadsCore : INotepadsCore
     {
@@ -798,13 +799,37 @@
         {
             if (Sets.Items?.Count > 1 && e.Set.Content is ITextEditor textEditor)
             {
-                // Only allow untitled empty document to be dragged outside for now
-                if (!textEditor.IsModified && textEditor.EditingFile == null)
-                {
-                    DeleteTextEditor(textEditor);
-                    await NotepadsProtocolService.LaunchProtocolAsync(NotepadsOperationProtocol.OpenNewInstance);
-                }
+                DeleteTextEditor(textEditor);
+                var message = new ValueSet();
+                message.Add("EditorData", JsonConvert.SerializeObject(await BuildTextEditorSessionData(textEditor), Formatting.Indented));
+                message.Add("LastSavedText", textEditor.IsModified && textEditor.EditingFile != null ? textEditor.LastSavedSnapshot.Content : null);
+                message.Add("PendingText", textEditor.IsModified ? textEditor.GetText() : null);
+                await NotepadsProtocolService.LaunchProtocolAsync(NotepadsOperationProtocol.OpenNewInstance, message);
             }
+        }
+
+        private async Task<TextEditorSessionDataV1> BuildTextEditorSessionData(ITextEditor textEditor)
+        {
+            TextEditorSessionDataV1 textEditorData = new TextEditorSessionDataV1
+            {
+                Id = textEditor.Id,
+            };
+
+            if (textEditor.EditingFile != null)
+            {
+                // Add the opened file to FutureAccessList so we can access it next launch
+                var futureAccessToken = textEditor.Id.ToString("N");
+                await FileSystemUtility.TryAddOrReplaceTokenInFutureAccessList(futureAccessToken, textEditor.EditingFile);
+                textEditorData.EditingFileFutureAccessToken = futureAccessToken;
+                textEditorData.EditingFileName = textEditor.EditingFileName;
+                textEditorData.EditingFilePath = textEditor.EditingFilePath;
+            }
+
+            textEditorData.PendingBackupFilePath = null;
+            textEditorData.LastSavedBackupFilePath = null;
+            textEditorData.StateMetaData = textEditor.GetTextEditorStateMetaData();
+
+            return textEditorData;
         }
 
         #endregion
--- a/src/Notepads/Views/NotepadsMainPage.IO.cs
+++ b/src/Notepads/Views/NotepadsMainPage.IO.cs
@@ -11,6 +11,8 @@
     using Notepads.Controls.TextEditor;
     using Notepads.Services;
     using Notepads.Utilities;
+    using Notepads.Models;
+    using Notepads.Core.SessionDataModels;
 
     public sealed partial class NotepadsMainPage
     {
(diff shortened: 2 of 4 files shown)


JasonStein:
Did you test if dragging a tab out when the file has been deleted or renamed? Try to see if it works.

soumyamahunt:
> Did you test if dragging a tab out when the file has been deleted or renamed? Try to see if it works.

I fixed it to work when filecontent is modified outside of app. For renamed/moved/deleted I am adding their content to the new file, can't think of a better solution. Anything on mind??

JasonStein:
> > Did you test if dragging a tab out when the file has been deleted or renamed? Try to see if it works.
> 
> I fixed it to work when filecontent is modified outside of app. For renamed/moved/deleted I am adding their content to the new file, can't think of a better solution. Anything on mind??

This is why I limiting empty file only.......... I could not make it work for deleted files.

JasonStein:
> > > Did you test if dragging a tab out when the file has been deleted or renamed? Try to see if it works.
> > 
> > 
> > I fixed it to work when filecontent is modified outside of app. For renamed/moved/deleted I am adding their content to the new file, can't think of a better solution. Anything on mind??
> 
> This is exactly why I am limiting it to empty file only.......... I could not make it work for deleted files scenarios.

If you cannot solve the problem, just disable the drag out in this case and re-pop the message to notify user that file has been moved or deleted.

soumyamahunt:
> If you cannot solve the problem, just disable the drag out in this case and re-pop the message to notify user that file has been moved or deleted (so he or she knows why dragging out of such tab won't work).

Fixed it.

src/Notepads/Core/NotepadsCore.cs
```diff
@@ -798,15 +799,46 @@ private async void Sets_SetDraggedOutside(object sender, SetDraggedOutsideEventA
         {
             if (Sets.Items?.Count > 1 && e.Set.Content is ITextEditor textEditor)
             {
-                // Only allow untitled empty document to be dragged outside for now
-                if (!textEditor.IsModified && textEditor.EditingFile == null)
+                if (textEditor.FileModificationState != FileModificationState.RenamedMovedOrDeleted)
                 {
                     DeleteTextEditor(textEditor);
```
JasonStein, line 804: Don't delete right away, the reason I put delete first was because the editor must be empty. Now we have value, so make we delete it after LaunchProtocolAsync returns true. You can also first make the eidtor.IsEnabled = false, and LaunchProtocolAsync, set it to true or deleted based on result.

src/Notepads/Views/NotepadsMainPage.IO.cs
```diff
@@ -210,5 +212,46 @@ private static bool HaveNonemptyTextEditor(ITextEditor[] textEditors)
             }
             return false;
         }
+
+        private async Task<ITextEditor> RecoverTextEditorAsync(TextEditorSessionDataV1 editorSessionData, string lastSavedText, string pendingText)
```
JasonStein, line 216: Make SessionManager.RecoverTextEditorAsync public and just reuse it.

src/Notepads/Core/NotepadsCore.cs
```diff
@@ -798,15 +799,46 @@ private async void Sets_SetDraggedOutside(object sender, SetDraggedOutsideEventA
...
-                    await NotepadsProtocolService.LaunchProtocolAsync(NotepadsOperationProtocol.OpenNewInstance);
+                    var message = new ValueSet();
+                    message.Add("EditorData", JsonConvert.SerializeObject(await BuildTextEditorSessionData(textEditor), Formatting.Indented));
+                    message.Add("LastSavedText", (textEditor.IsModified || textEditor.FileModificationState == FileModificationState.Modified) && textEditor.EditingFile != null ? textEditor.LastSavedSnapshot.Content : null);
+                    message.Add("PendingText", textEditor.IsModified ? textEditor.GetText() : null);
```
JasonStein, line 808: Did you verify the upper size limit of message when transferring using the launch protocol? Try test with a large file to see if it works.

JasonStein: > Did you verify the upper size limit of message when transferring using the launch protocol? Try test with a large file to see if it works.

This is another reason I did not implement the dragging for non empty eidtors.

soumyamahunt: Made the approach similar to session snapshot, now it will work.

JasonStein:
Another thing is that when you drag a tab outside to a desktop for example, it shows a "move" menu instead of creating the new instance directly and you have to manually cancel the menu before it can work. This needs to be solved otherwise the UX will be very confusing.

soumyamahunt:
> Another thing is that when you drag a tab outside to a desktop for example, it shows a "move" menu instead of creating the new instance directly and you have to manually cancel the menu before it can work. This needs to be solved otherwise the UX will be very confusing.

Fixed this, however some apps like VS Code and Firefox still able to intercept drop action even if the tab doesn't contain any file. Also, some downside of this fix is no more drag and drop sharing between notepads and other programs.

JasonStein:
> > Another thing is that when you drag a tab outside to a desktop for example, it shows a "move" menu instead of creating the new instance directly and you have to manually cancel the menu before it can work. This needs to be solved otherwise the UX will be very confusing.
> 
> Fixed this, however some apps like VS Code and Firefox still able to intercept drop action even if the tab doesn't contain any file. Also, some downside of this fix is no more drag and drop sharing between notepads and other programs may be implementing "sharing" or "open with another app" features will be the way to go.

If this is going to break the sharing between notepads then I would say don't do it.... 

Another approach I think is better:

1. Keep everything as it is: 1) User can still drag out empty tabs. 2) Dragging out tab which has a file to third party apps will still work. 3) Dragging between notepads is un-touched.

2. Add one more menu item in TabContextFlyout and call it "Open in new window". Put 100% of your logic/changes there.

soumyamahunt:
> 1. Keep everything as it is: 1) User can still drag out empty tabs. 2) Dragging out tab which has a file to third party apps will still work. 3) Dragging between notepads is un-touched.
> 2. Add one more menu item in TabContextFlyout and call it "Open in new window". Put 100% of your logic/changes there.

I went for the opposite approach. I kept my implementation of tab drag and drop. To share files to other apps via drag and drop user can drag the PathIndicator to the other app. This will clear any ux confusion. Is this acceptable??

JasonStein:
> > 1. Keep everything as it is: 1) User can still drag out empty tabs. 2) Dragging out tab which has a file to third party apps will still work. 3) Dragging between notepads is un-touched.
> > 2. Add one more menu item in TabContextFlyout and call it "Open in new window". Put 100% of your logic/changes there.
> 
> I went for the opposite approach. I kept my implementation of tab drag and drop. To share files to other apps via drag and drop user can drag the PathIndicator to the other app. This will clear any ux confusion. Is this acceptable??

Still feeling awkward since this is not a well established UX behavior. Since WinUI TabView will eventually abandon the ListView based TabView and will use ItemRepeater for proper implementation. So we will be able to use it as well. At the time, we can get rid of the annoying drag n drop behavior brought by the ListViewBase. Once that happens, we can just point it to the code you have written. That's also why I suggest you to go with "Open in new window" route and keep everything as it is.

soumyamahunt:
> Still feeling awkward since this is not a well established UX behavior. 

The typical behavior of dragging tab outside of an app is to create new window of the app. This seems to be standard across  various tabbed interfaces like browsers and editors. That's why I think implementing file transfer through setsitem drag-drop will confuse user.

>Since WinUI TabView will eventually abandon the ListView based TabView and will use ItemRepeater for proper implementation. So we will be able to use it as well. At the time, we can get rid of the annoying drag n drop behavior brought by the ListViewBase. Once that happens, we can just point it to the code you have written. That's also why I suggest you to go with "Open in new window" route and keep everything as it is.

If you don't want to implement this now I would suggest to keep it open until WinUI 3 releases. Currently users can create a new window and drag the instance to that window I don't think cluttering the TabContextFlyout is necessary for this.

soumyamahunt:
@JasonStein I have made this behavior consistent for tabs for which files are either renamed or moved. Hope this removes any kind of confusion. You can implement this whenever you want.

wuhFS9nDiM

LiBp3I8Sjm

PR #490, opened 2020-05-04
Linehighlighter bug-fix

soumyamahunt:
- Fixed bug when highlighter height wasn't changing when some text is selected and font size changed.

- Highlighter height is now more accurate.

## PR Type

Bugfix


JasonStein:
Thanks!

PR #495, opened 2020-05-05
ScrollBar context menu.

soumyamahunt:
This PR implements notepad style scrollbar context menu but has some issues, @JasonStein can you help with this??

## PR Type

Feature

## Other information

The ScrollViewer resets instantly after changing, I can't pinpoint the cause.

1HOoXiWJ93

### DIFF
--- a/src/Notepads/Controls/TextEditor/TextEditorCore.cs
+++ b/src/Notepads/Controls/TextEditor/TextEditorCore.cs
@@ -66,8 +66,12 @@
         private Canvas _lineNumberCanvas;
         private const string LineNumberGridName = "LineNumberGrid";
         private Grid _lineNumberGrid;
+        private const string ContentScrollViewerHorizontalScrollBarName = "HorizontalScrollBar";
+        private ScrollBar _contentScrollViewerHorizontalScrollBar;
+        private TextEditorScrollBarContextFlyout _contentScrollViewerHorizontalScrollBarFlyout;
         private const string ContentScrollViewerVerticalScrollBarName = "VerticalScrollBar";
         private ScrollBar _contentScrollViewerVerticalScrollBar;
+        private TextEditorScrollBarContextFlyout _contentScrollViewerVerticalScrollBarFlyout;
         private const string LineHighlighterAndIndicatorCanvasName = "LineHighlighterAndIndicatorCanvas";
         private Canvas _lineHighlighterAndIndicatorCanvas;
         private const string LineHighlighterName = "LineHighlighter";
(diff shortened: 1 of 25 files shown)


JasonStein:
Before you do change view, set this variable to false:
_shouldResetScrollViewerToLastKnownPositionAfterRegainingFocus = false;

Also, you can consider change selection point to 0 after user hit scroll to top button and change selection point to end of the document after user hit scroll to bottom.

soumyamahunt:
> Before you do change view, set this variable to false:
> _shouldResetScrollViewerToLastKnownPositionAfterRegainingFocus = false;

This doesn't work, also I am changing previous offsets before changing scroll position so it wouldn't be necessary.

> Also, you can consider change selection point to 0 after user hit scroll to top button and change selection point to end of the document after user hit scroll to bottom.

Did this.

soumyamahunt:
Also there is another bug, after clicking any entry from flyout when I click on any where on the document it immitates click and hold.

qhdkvsVADt

JasonStein:
This is why it is not going to work:

image

When the context flyout is open, editor lost its focus, then you change the view right away and editor regain focus. Then editor itself change it's view to selection point (this is done by the internal logic which cannot be modified) and then it screws up the whole thing. 

Let me work on it, I think I might be able to find a workaround.

JasonStein:
I think I fixed the issue but still, there are two bugs that I don't even think is solvable:

1. Right now the vertical context flyout works great but not for the horizontal one. I don't think it is our fault, it should be a bug related to RichEdit/ScrollViewer.

2. After changing the view, if you click on any of the text and move your mouse, it will select from there acting like you are doing mouse click + move. This is super wired and I do not know why and how to fix it.

This is as far as I can go, you can take from here.

JasonStein:
@soumyamahunt  Any luck? Feel free to spin up another bug report to WinUI repo :)

soumyamahunt:
> @soumyamahunt Any luck? Feel free to spin up another bug report to WinUI repo :)

Filed [here](https://github.com/microsoft/microsoft-ui-xaml/issues/2415).

JasonStein:
@soumyamahunt Give this a try? https://docs.microsoft.com/en-us/uwp/api/windows.ui.text.itextrange.scrollintoview?view=winrt-19041#Windows_UI_Text_ITextRange_ScrollIntoView_Windows_UI_Text_PointOptions_

soumyamahunt:
> @soumyamahunt Give this a try? https://docs.microsoft.com/en-us/uwp/api/windows.ui.text.itextrange.scrollintoview?view=winrt-19041#Windows_UI_Text_ITextRange_ScrollIntoView_Windows_UI_Text_PointOptions_

This fixes horizontal scroll issue. However this issue not fixed:

> 2. After changing the view, if you click on any of the text and move your mouse, it will select from there acting like you are doing mouse left click + text selection. This is super wired and I do not know why and how to fix it.

JasonStein:
> > @soumyamahunt Give this a try? https://docs.microsoft.com/en-us/uwp/api/windows.ui.text.itextrange.scrollintoview?view=winrt-19041#Windows_UI_Text_ITextRange_ScrollIntoView_Windows_UI_Text_PointOptions_
> 
> This fixes horizontal scroll issue. However this issue not fixed:
> 
> > 1. After changing the view, if you click on any of the text and move your mouse, it will select from there acting like you are doing mouse left click + text selection. This is super wired and I do not know why and how to fix it.

Did you try removing the logic of doing two manual ChangeView events on the ScrollViewer that I did?

soumyamahunt:
> Did you try removing the logic of doing two manual ChangeView events on the ScrollViewer that I did?

Removing those makes the scrolling not work. Scrolling is again being reset after gaining focus.

JasonStein:
> > Did you try removing the logic of doing two manual ChangeView events on the ScrollViewer that I did?
> 
> Removing those makes the scrolling not work. Scrolling is again being reset after gaining focus.

What about doing ITextRange ScrollIntoView after getting focus instead of doing it right away before ChangeView?

soumyamahunt:
> What about doing ITextRange ScrollIntoView after getting focus instead of doing it right away before ChangeView?

Doesn't work this way either.

JasonStein:
> > What about doing ITextRange ScrollIntoView after getting focus instead of doing it right away before ChangeView?
> 
> Doesn't work this way either.

Ok, let me give it a try.

PR #502, opened 2020-05-08
Fix the drag and drop saves

Maickonn:
Fix the "If you drag a file into Notepads, file save picker will ask you to save it before closing due to UWP restriction."

## PR Type
Bugfix

## Other information
**For correct operation, follow these steps:**
No need broadFileSystemAccess permission.

### DIFF
--- a/src/Notepads/Views/NotepadsMainPage.IO.cs
+++ b/src/Notepads/Views/NotepadsMainPage.IO.cs
@@ -108,6 +108,30 @@
             return successCount;
         }
 
+        private async Task<StorageFile> OpenSaveAs(ITextEditor textEditor)
+        {
+            NotepadsCore.SwitchTo(textEditor);
+            StorageFile file = await FilePickerFactory.GetFileSavePicker(textEditor, true).PickSaveFileAsync();
+            NotepadsCore.FocusOnTextEditor(textEditor);
+            return file;
+        }
+
+        private async Task TrySave(ITextEditor textEditor, StorageFile file, bool rebuildOpenRecentItems)
+        {
+            try
+            {
+                await NotepadsCore.SaveContentToFileAndUpdateEditorState(textEditor, file);
+                var success = MRUService.TryAdd(file); // Remember recently used files
+                if (success && rebuildOpenRecentItems)
+                {
+                    await BuildOpenRecentButtonSubItems();
+                }
+            } catch (Exception)
+            {
+                throw new UnauthorizedAccessException("Access denied");
+            }
+        }
+
         private async Task<bool> Save(ITextEditor textEditor, bool saveAs, bool ignoreUnmodifiedDocument = false, bool rebuildOpenRecentItems = true)
         {
             if (textEditor == null) return false;
@@ -122,9 +146,7 @@
             {
                 if (textEditor.EditingFile == null || saveAs)
                 {
-                    NotepadsCore.SwitchTo(textEditor);
-                    file = await FilePickerFactory.GetFileSavePicker(textEditor, saveAs).PickSaveFileAsync();
-                    NotepadsCore.FocusOnTextEditor(textEditor);
+                    file = await OpenSaveAs(textEditor);
                     if (file == null)
                     {
                         return false; // User cancelled
@@ -135,12 +157,18 @@
                     file = textEditor.EditingFile;
                 }
 
-                await NotepadsCore.SaveContentToFileAndUpdateEditorState(textEditor, file);
-                var success = MRUService.TryAdd(file); // Remember recently used files
-                if (success && rebuildOpenRecentItems)
+                await TrySave(textEditor, file, rebuildOpenRecentItems);
+                return true;
+            }
+            catch(UnauthorizedAccessException ex)
+            {
+                file = await OpenSaveAs(textEditor);
+                if (file == null)
                 {
-                    await BuildOpenRecentButtonSubItems();
+                    return false; // User cancelled
                 }
+
+                await TrySave(textEditor, file, rebuildOpenRecentItems);
                 return true;
             }
             catch (Exception ex)
--- a/src/Notepads/Views/NotepadsMainPage.IO.cs
+++ b/src/Notepads/Views/NotepadsMainPage.IO.cs
@@ -11,6 +11,7 @@
     using Notepads.Controls.TextEditor;
     using Notepads.Services;
     using Notepads.Utilities;
+    using System.IO;
 
     public sealed partial class NotepadsMainPage
     {
@@ -128,7 +129,7 @@
                 }
             } catch (Exception)
             {
-                throw new UnauthorizedAccessException("Access denied");
+                throw new IOException();
             }
         }
 
@@ -160,7 +161,7 @@
                 await TrySave(textEditor, file, rebuildOpenRecentItems);
                 return true;
             }
-            catch(UnauthorizedAccessException ex)
+            catch(IOException)
             {
                 file = await OpenSaveAs(textEditor);
                 if (file == null)
--- a/src/Notepads/Views/NotepadsMainPage.IO.cs
+++ b/src/Notepads/Views/NotepadsMainPage.IO.cs
@@ -56,6 +56,10 @@
             try
             {
                 if (file == null) return false;
+                
+                if (!await FileSystemUtility.FileIsWritable(file))
+                    file = await FileSystemUtility.GetFile(file.Path);
+
                 var openedEditor = NotepadsCore.GetTextEditor(file);
                 if (openedEditor != null)
                 {
--- a/src/Notepads/Views/NotepadsMainPage.IO.cs
+++ b/src/Notepads/Views/NotepadsMainPage.IO.cs
@@ -56,10 +56,6 @@
             try
             {
                 if (file == null) return false;
-                
-                if (!await FileSystemUtility.FileIsWritable(file))
-                    file = await FileSystemUtility.GetFile(file.Path);
-
                 var openedEditor = NotepadsCore.GetTextEditor(file);
                 if (openedEditor != null)
                 {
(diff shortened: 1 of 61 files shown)


JasonStein:
Take a look at this one: https://github.com/JasonStein/Notepads/pull/364

Maickonn:
> Take a look at this one: #364

lol, I had not seen this issue. I believe there is another way to correct this. I will implement here.

Maickonn:
@JasonStein , It's done!
But I couldn't find a way to pass encoding as an argument.

JasonStein:
> @JasonStein , It's done!
> But I couldn't find a way to pass encoding as an argument.

That's exactly what I have discovered. I knew this "PathIO" trick last week but no, it is not going to work for Notepads scenario.

Maickonn:
> > @JasonStein , It's done!
> > But I couldn't find a way to pass encoding as an argument.
> 
> That's exactly what I have discovered. I knew this "PathIO" trick last week but no, it is not going to work for Notepads scenario.

Ok. Can you explain the reason?

JasonStein:
> > > @JasonStein , It's done!
> > > But I couldn't find a way to pass encoding as an argument.
> > 
> > 
> > That's exactly what I have discovered. I knew this "PathIO" trick last week but no, it is not going to work for Notepads scenario.
> 
> Ok. Can you explain the reason?

Try put in the encoding first: PathIO.WriteTextAsync(file.Path, text, encoding);

Maickonn:
Okay.

Maickonn:
@JasonStein , It's done.

src/Notepads/Views/NotepadsMainPage.IO.cs
```diff
@@ -120,9 +120,7 @@ private async Task<bool> Save(ITextEditor textEditor, bool saveAs, bool ignoreUn
             StorageFile file = null;
             try
             {
-                if (textEditor.EditingFile == null || saveAs ||
-                    FileSystemUtility.IsFileReadOnly(textEditor.EditingFile) ||
```
JasonStein, line 124: FileSystemUtility.IsFileReadOnly(textEditor.EditingFile)
You need to keep this because for files those are actually read-only, it should prompt save picker.

JasonStein: Another question for you is that what if the file been dragged into notepads is actually a read-only file?

Maickonn: I did it because as you said, when dragging a file into Notepads, it automatically becomes a "fake" Read Only, so it would not enter the save file function, where it has the trick to check if the file is really Ready Only.

When a file is a real Read Only and the user tries to save it, Notepads automatically displays a message stating that it cannot be saved to it. I believe that this way it becomes more explanatory for the user to know the reason for not being able to overwrite.
Capture

JasonStein: Not true since:

1. This is not obvious to user, since this is an error popup.
2. People don't read exception message.
3. "Access is denied" only makes sense to developers not general audience. I am pretty sure some of our users will not map it to "File is read only, please do save as"

That is also why I am popping up save dialog when this happens.

Maickonn: Okay. I will implement the prompt Save As when save fails.

JasonStein: > Okay. I will implement the prompt Save As when save fails.

I personally against using Path.IO since this is a kind of "expliot" to me.

But if you really want to do it, here is what I would suggest:

1. Whenever file is been dragged into notepads, you somehow mark it as "OpenedByDragging".

2. You check this flag explicitly when Save triggers. If you don't see this flag. Just follow the exciting code path and don't touch it.

3. If you are seeing this flag. Do below steps:
    1) Check if file is writable, if it is, then we are happy and you can do Path.IO.
    2) If it is not writable, fallback to open the save picker. (This is the case where the file was read-only before dragging into notepads)

Maickonn: @JasonStein 

I was carrying out the implementation according to your guidelines, but in the course of it, I found that there is no way to know if the file is really writable without trying to write to it.

I believe that the best way to do this is to control exceptions. For example, if you are unable to write, throw an exception to open File Save Picker. What do you think?

JasonStein: I don't quite like that approach tho (also exception message varies between different locales), let me take a look first. Also, what about the case when you drag a file into Notepads and then exit the app and reopen and then do save?

Maickonn: I'm reopening the pull request for you to look at the changes.

If you look at the commits tree, you will notice that I was implementing "isFileOpenedByDragging", but as there was no way to check if the file is a real read only without writing to it, I chose to abandon it and try the exception path.

Watch this demo video
[](https://youtu.be/mppWY_1wGRA)
https://youtu.be/mppWY_1wGRA

jpaulobatista: I tested it here and it worked. Good job Maickonn

JasonStein: @Maickonn  You can remove all the changes related to "isFileOpenedByDragging'. I will test other scenarios in the meantime.

Maickonn: > @Maickonn You can remove all the changes related to "isFileOpenedByDragging'. I will test other scenarios in the meantime.

Okay.

Maickonn: @JasonStein it's done.

src/Notepads/Views/NotepadsMainPage.IO.cs
```diff
@@ -108,6 +108,30 @@ public async Task<int> OpenFiles(IReadOnlyList<IStorageItem> storageItems)
...
+            {
+                await NotepadsCore.SaveContentToFileAndUpdateEditorState(textEditor, file);
+                var success = MRUService.TryAdd(file); // Remember recently used files
+                if (success && rebuildOpenRecentItems)
+                {
+                    await BuildOpenRecentButtonSubItems();
+                }
+            } catch (Exception)
+            {
+                throw new UnauthorizedAccessException("Access denied");
```
JasonStein, line 131: There are other cases where you might get this exception. Like the drive could be removed or the cloud drive connection lost. You cannot assume it is access denied here.

src/Notepads/Views/NotepadsMainPage.IO.cs
JasonStein: On top of what you have tested, there are few more cases you need to test as well:

1. Open file from SMB share and disconnect, then try to save.
2. Open file from CloudDrive like OneDrive or Dropbox, disconnect it and then try to save.
3. Open file from a USB drive, disconnect it and then try to save.
4. Make sure session snapshot is enabled then you open the file, do some changes, close the app and reopen the app. Now try to save.

We need to make sure all above cases work as expected and show proper message to the user accordingly.

Maickonn: > On top of what you have tested, there are few more cases you need to test as well:
> 
> 1. Open file from SMB share and disconnect, then try to save.
> 2. Open file from CloudDrive like OneDrive or Dropbox, disconnect it and then try to save.
> 3. Open file from a USB drive, disconnect it and then try to save.
> 4. Make sure session snapshot is enabled then you open the file, do some changes, close the app and reopen the app. Now try to save.
> 
> We need to make sure all above cases work as expected and show proper message to the user accordingly.

As requested, I tested on all listed scenarios. When the file cannot be written or is no longer accessible, Notepads opens a Save As.

About session snapshot, it is working correctly, even when the file has been opened by drag and drop.

src/Notepads/Views/NotepadsMainPage.IO.cs
Maickonn: > There are other cases where you might get this exception. Like the drive could be removed or the cloud drive connection lost. You cannot assume it is access denied here.

I changed from UnauthorizedAccessException to IOException.

According to the documentation, the exception that is thrown when an I/O error occurs.

I think she fits the case. That is, if you are unable to save the file for any reason, open a Save As.

And when you use FileSavePicker and try to save to a location that is blocked by the system, windows automatically denies it stating the reason (if the location is blocked, it doesn't exist, real read only, etc).

Untitled Project

JasonStein: There are still few potential problems with this implementation. Here is mine based on yours: https://github.com/JasonStein/Notepads/pull/512/files
Can you take a look?

JasonStein: Few problems I solved this my implementation:

1. It properly saves UTF-8-BOM or any encoding that has a preamble.
2. Call CompleteUpdatesAsync in case exception happens.
3. No self created exception anymore since I found that the exception type is actually "UnauthorizedAccessException" when file is read-only and you are trying to save.

JasonStein: If that PR looks good to you, please do following:

1. Abandon this one and create a new one with my changes and send out a PR.
2. I will close mine and accept yours so you get the credits.
3. Before you do so, please help me test all scenarios again to make sure it works in all cases.

Another test case would be editing a file and save it while OneDrive/Dropbox is updating it. Not sure what will happen in that case but it should be handled correctly.

Maickonn: I merged yours implementations, I performed the tests again and I fixed an problem when you tried save file not found or on storage removed, he showed the error dialog. Then I added "FileNotFoundException" on catch to solve it.

I performed tests in several scenarios, such as Dropbox while updating the file, removing external storage, deleting file in editing and deleting file in shared network edition.

JasonStein: LGTM, thanks for the effort!

PR #510, opened 2020-05-09
Compressed images

VeryRealBot:
Images that are futher compressed than what was done by imgbot to reduce the file size of this application


JasonStein:
Quality lose is not acceptable: 
Screen Shot 2020-05-08 at 6 03 00 PM

PR #514, opened 2020-05-10
Translation/updated PT-BR

pinholucas:
New string in manifest file and maybe a better translation to "Print All".

## PR Type

- Translation


pinholucas:
Wtf, my changes in Manifest file wasn't uploaded

PR #523, opened 2020-05-11
Implemented desktop extension to modify system file

soumyamahunt:
Implemented desktop extension to modify system files directly. User needs to provide admin access to Notepads to do so.

## PR Type

Feature

## Other information
UwDfmbD7dN


soumyamahunt:
@JasonStein I just noticed after writing to system file multiple times, for some cases you don't need admin permission anymore. The storage api is really flakey.

JasonStein:
> @JasonStein I just noticed after writing to system file multiple times, for some cases you don't need admin permission anymore. The storage api is really flakey.

WHAT? srsly... You should re-pro it and send another bug report to WinUI repo....

soumyamahunt:
> WHAT? srsly... You should re-pro it and send another bug report to WinUI repo....

Here is an example:
jcZ8fpIiYl

I fear rather than solving it they would totally remove elevation ability.

JasonStein:
> > WHAT? srsly... You should re-pro it and send another bug report to WinUI repo....
> 
> Here is an example:
> jcZ8fpIiYl
> 
> I fear rather than solving it they would totally remove elevation ability. By the way, this is the only file in my testing that produces this issue so far.

Should still report it tho :)

src/Notepads/Services/DesktopExtensionService.cs
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        public static async Task SaveFileAsAdmin(string filePath, byte[] data)
+        {
+            if (InteropServiceConnection == null) await Initialize();
+
+            using (var adminConnectionPipeStream = new NamedPipeServerStream($"Local\\{Package.Current.Id.FamilyName}\\AdminWritePipe",
+                PipeDirection.InOut, 254, PipeTransmissionMode.Message, PipeOptions.Asynchronous))
+            {
+                if (!adminConnectionPipeStream.WaitForConnectionAsync().Wait(250))
```
soumyamahunt, line 79: @JasonStein what could be the minimum ms we can wait here??

JasonStein: Is it going to affect on startup?

soumyamahunt: No, this would only apply when user is trying to modify a system file. Notepads tries to connect to the desktop extension component within 250ms. If it succeeds it sends the data to the component otherwise it prompts user to run the extension as adminstrator.

soumyamahunt:
@JasonStein I am converting this pr into draft cause the extension is based on dotnet 5. After dotnet 5 becomes stable I will make this pr available to merge..

src/Notepads/Program.cs
```diff
@@ -35,13 +38,14 @@ static void Main(string[] args)
             {
                 IsPrimaryInstance = true;
                 ApplicationSettingsStore.Write(SettingsKey.ActiveInstanceIdStr, null);
-                instanceHandlerMutex.ReleaseMutex();
```
soumyamahunt, line 38: @JasonStein plz make this change in main branch too. When compiled in "Release" mode this line causes some issues, removing this should fix that.

JasonStein: What will happen if I remove this?

soumyamahunt: Currently when launching more instances, new instance might not detect the primary instance is present and set itself as primary. But removing this line fixes this. Also, this issue only happens only when app is compiled in release mode.

soumyamahunt:
@JasonStein Since .NET 5 is released now and work on implementing this feature is completed in my opinion, I am making this PR ready to merge.

soumyamahunt:
@JasonStein using .NET5 for desktop extension increased the deployment size significantly (even after trimming), so I decided to move the desktop extension project to C++.

PR #540, opened 2020-05-15
Implemented optional language change feature

soumyamahunt:
Implemented optional language change feature for Internationalization and Localization contributors. Contributors don't have to sign out to change the App's language. Changing language inside app and restarting changes App's language.

## PR Type

Feature


JasonStein:
Great, I am ok with DEBUG only. Thanks!

PR #541, opened 2020-05-16
Removed shadows on 24-px icons

mah-qurashy:
When creating the new icons, i intentionally added shadows on the 24-px icon to have better visibility on the light mode taskbar
now i noticed that the dark icon is actually used on the light mode taskbar
image
not sure why this wasn't the case before, is that something that was recently committed? @JasonStein. or maybe something prevented windows from updating the icon when i changed color mode on my machine.
either way, i removed the shadow as now it serves no purpose and to keep all sizes consistent


JasonStein:
It has always been like that, dark icon on light taskbar and light icon on dark taskbar.

PR #557, opened 2020-05-25
Fix for moving lines with Alt+Up/Down

soumyamahunt:
Fixed moving lines with Alt+Up/Down.

## PR Type

Bugfix


JasonStein:
Another bug I found:

image
Now, do alt+down twice and then alt+up

JasonStein:
Also found an issue in indentation:
Before:
image

After:
image
(last line is gone)

soumyamahunt:
@JasonStein fixed all the issues reported so far.

PR #564, opened 2020-05-28
Hook up foreground operations when running as a Game Bar widget

totallymonkey:
This change updates how foreground operations were being done in Notepads when running as a Game Bar widget. I created a new wrapper object that encapsulates the complicated logic for actually calling the Game Bar API. All that is needed is to provide a function delegate that runs the UI thread code. In addition, there is an event that needs to be listened to off of the wrapper object that will inform the caller that the operation should be canceled.

### DIFF
--- a/src/Notepads.Controls/Notepads.Controls.csproj
+++ b/src/Notepads.Controls/Notepads.Controls.csproj
@@ -235,7 +235,7 @@
       <Version>2.0.6</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform">
-      <Version>6.2.10</Version>
+      <Version>6.2.11-preview-28914-04</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.Toolkit.Uwp.UI">
       <Version>6.0.0</Version>
--- a/src/Notepads.Controls/Notepads.Controls.csproj
+++ b/src/Notepads.Controls/Notepads.Controls.csproj
@@ -235,7 +235,7 @@
       <Version>2.0.6</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform">
-      <Version>6.2.11-preview-28914-04</Version>
+      <Version>6.2.10</Version>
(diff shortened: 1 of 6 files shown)


src/Notepads/Notepads.csproj
```diff
@@ -225,6 +225,7 @@
     <Compile Include="Models\TextFile.cs" />
     <Compile Include="Extensions\StringExtensions.cs" />
     <Compile Include="Utilities\ThreadUtility.cs" />
+    <Compile Include="XboxGameBarForegroundWorkerWrapper.cs" />
```
totallymonkey, line 228: I wasn't sure exactly where you would want this file, it didn't seem to fit cleanly in any of your existing folders. Let me know what you think.

src/Notepads.Controls/Notepads.Controls.csproj
```diff
@@ -235,7 +235,7 @@
       <Version>2.0.6</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform">
-      <Version>6.2.10</Version>
+      <Version>6.2.11-preview-28914-04</Version>
```
vidager, line 238: Should this be changed?

src/Notepads/Notepads.csproj
JasonStein: You can create a Widget folder and put xbox code/class there. Maybe also create a IForegroundWorkerWrapper interface and have a default implementation for handling normal (non-wdiget) scenario (ForegroundWorkerWrapper.cs). 

Based on App.IsGameBarWdiget and _widget != null, you can then create a XboxGameBarForegroundWorkerWrapper class on the fly to handle widget file opening/saving or alternatively, use ForegroundWorkerWrapper for normal path. Otherwise even for normal Notepads instance, game bar DLLs will be loaded when opening file save/open dialogs.

src/Notepads.Controls/Notepads.Controls.csproj
JasonStein: Yeah, revert it and I can handle it offline.

PR #568, opened 2020-05-30
Updated Readme with New Shortcut Description

subodhpareek18:
## PR Type

Documentation content changes


JasonStein:
I don't want to add these kind of shortcut in the readme because they are well known and used by other text editor or code editor as well. Same for "Ctrl+D" to replicate selections etc.

PR #576, opened 2020-06-05
Added moving text with Alt+Left/Right

soumyamahunt:
Added moving text with Alt+Left/Right arrrow.

## PR Type

Feature
## Other information
I found none of the windows' text editor has this feature but gnome-editor in Linux has this behavior so I added this to Notepads.
EwhtKXupjD
9h4kc6rnJO
32iFd3R077
kWyEcrsQDv
XtTHlqoSQt


JasonStein:
There are some code used in MoveTextRight and Left can be move out of them for sharing. Can you make some refactoring here to reduce the length/complexity of the method?

soumyamahunt:
@JasonStein reduced complexity in the methods.

JasonStein:
Not sure if this is intended:
image
Click alt+right two times then do alt+left.

soumyamahunt:
> @soumyamahunt
> Not sure if this is intended:
> image
> Click alt+right two times then do alt+left.

Yes this is intended behavior.

PR #590, opened 2020-06-13
Potential fix to show all supported languages

soumyamahunt:
Fix to show all supported languages in language options.

## PR Type

Bugfix

### DIFF
--- a/src/Notepads/Utilities/LanguageUtility.cs
+++ b/src/Notepads/Utilities/LanguageUtility.cs
@@ -27,13 +27,39 @@
 
     public static class LanguageUtility
     {
+        private static readonly IReadOnlyList<string> _manifestLanguages = new List<string>()
+        {
+            "en-US",
+            "bg",
+            "cs", 
+            "de-CH",
+            "de-DE",
+            "es-ES", 
+            "fi",
+            "fr-FR", 
+            "hi", 
+            "hr-HR", 
+            "hu",
+            "it-IT",
+            "ja",
+            "ka",
+            "ko",
+            "or",
+            "pl",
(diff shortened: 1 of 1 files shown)


soumyamahunt:
@JasonStein can you test if installing from package shows all the manifest languages??

JasonStein:
It works locally in release mode but my guess is that it will not work once built and distributed by the store. But that just be my guess.

https://docs.microsoft.com/en-us/windows/uwp/design/globalizing/manage-language-and-region
https://developercommunity.visualstudio.com/content/problem/1041637/the-new-version-probably-vs-makes-the-language-lis.html

Maybe we can use this:

```csharp
    IReadOnlyList<string> runtimeLanguages = Windows.ApplicationModel.Resources.Core.ResourceContext.GetForCurrentView().Languages;
    runtimeLanguages = Windows.Globalization.ApplicationLanguages.Languages;
```

soumyamahunt:
> Maybe we can use this:
> 
> ```cs
>     IReadOnlyList<string> runtimeLanguages = Windows.ApplicationModel.Resources.Core.ResourceContext.GetForCurrentView().Languages;
>     runtimeLanguages = Windows.Globalization.ApplicationLanguages.Languages;
> ```

These methods will only show languages installed by the user.

> It works locally in release mode but my guess is that it will not work once built and distributed by the store. But that just be my guess.

[This](https://www.guruumeditation.net/dynamically-change-the-language-of-a-universal-app/) article describes the problem it doesn't work in store package because during package creation value of “Create App bundle” isn't set to "Never". May be changing that will solve the issue??

Issue #248, opened 2019-11-21
[Feature request] Exiting the app, when the last tab is closed


sydeslyde:
**Is your feature request related to a problem? Please describe.**
Not really a problem. I think of it as a usability/customisation improvement.

**Describe the solution you'd like**
I would like the app to exit, when the last open tab is closed. There could be a setting (in the Advanced Settings), to let the user decide if the app should behave this way.
Notepad++ has a setting called "Exit on close the last tab" for this. I don't quite like the wording, I would call it "Exit app when last tab is closed".

**Describe alternatives you've considered**
-

**Additional context**
I have already forked the repo and implemented the described solution. If you want, I would be happy to create a pull request.
![notepads_screenshot](https://user-images.githubusercontent.com/34600077/69350295-9ba3c980-0c79-11ea-9091-1d5116894c7e.png)

JasonStein:
The main issue is that Windows will not remember window last opening position and size if the app is exited programmatically. I have OCD and I do not want that to happen.

soumyamahunt:
@sydeslyde implemented in #601.

PR #601, opened 2020-06-24
feat: exit app when closing last tab

soumyamahunt:
Implemented functionality to exit app when last tab is closed as requested in #248.

## PR Type

- Feature

## Other information
The implemented method doesn't interfere with Windows' ability to remember window size and position.

W13JDdBKCD

M3mIjjDjVO


JasonStein:
Never realized there is a method called "TryConsolidateAsync" and does exactly what is needed. How did you figure that out?

soumyamahunt:
> Never realized there is a method called "TryConsolidateAsync" and does exactly what is needed. How did you figure that out?

Just by reading documentation.

JasonStein:
Since it has been too long that I changed this behavior (it used to be like this). Maybe we should add this as an option in Advanced Settings? Default would be not to close the app and let's give the option to user.

soumyamahunt:
> Since it has been too long that I changed this behavior (it used to be like this). Maybe we should add this as an option in Advanced Settings? Default would be not to close the app and let's give the option to user.

Made changes as you asked, here is how the ui looks:

ApplicationFrameHost_NvKe4ul4Ys

sydeslyde:
I had a go at that myself, but didn't know about `TryConsolidateAsync()`, very nice.
+1 for merging this.
I'll happily supply German translations when this gets to master.

soumyamahunt:
@JasonStein is there any further work you need from me here or is it RTM??

JasonStein:
> @JasonStein is there any further work you need from me here or is it RTM??

image
I don't like the translation here since it is not very user-friendly. Let me think of any better wording here before merging the PR. Btw I will probably put this in the next release.

sydeslyde:
Some Suggestions for the wording:

- "App Behavior Preferences"
- "Window Behavior Preferences"
- "Exit app when closing last tab"

JasonStein:
> * "Window Behavior Preferences"
> * "Exit app when closing last tab"

LGTM. @soumyamahunt thoughts?

soumyamahunt:
> > * "Window Behavior Preferences"
> > * "Exit app when closing last tab"
> 
> LGTM. @soumyamahunt thoughts?

Fine with this, but I think **"Always open new window"** conveys better meaning than **"Window Behavior Preferences"**.

soumyamahunt:
This is how it looks now after the changes:

ApplicationFrameHost_lBRJmctwa3

sydeslyde:
The description below the switches is confusing. Its a single sentence, describing two independent settings. When I read it, it appears to me, that "closing the last tab" will also somehow enable "always open in new window".

I think it should be split up to describe the options seperately, or maybe we should just omit the description of "Exit app when closing last tab", since the title explains it well enough.

soumyamahunt:
> The description below the switches is confusing. Its a single sentence, describing two independent settings. When I read it, it appears to me, that "closing the last tab" will also somehow enable "always open in new window".
> 
> I think it should be split up to describe the options seperately, or maybe we should just omit the description of "Exit app when closing last tab", since the title explains it well enough.

Yes this could create confusion, I have removed description for last tab closing option.

ApplicationFrameHost_2KwZd6pcKg

UmbralReaper:
What about moving the description to beneath the first switch, and have the description change depending on whether the switch is on or off. If the switch is on it reads `The app is automatically closed when there are no open tabs.`. If the switch is off it reads `The app stays open when there are no open tabs.`

soumyamahunt:
> What about moving the description to beneath the first switch, and have the description change depending on whether the switch is on or off. If the switch is on it reads `The app is automatically closed when there are no open tabs.`. If the switch is off it reads `The app stays open when there are no open tabs.`

Notepads doesn't follow different content depending on whether switch is on or off. Doing that only for this will create confusion. Also, I personally think this type of changing content whether switch is on/off creates more confusion for user. In your example when the user toggles it on it reads `The app is automatically closed when there are no open tabs.` does that mean currently this behavior applied or when user toggles switch off the behavior is applied, which creates confusion.

UmbralReaper:
> Notepads doesn't follow different content depending on whether switch is on or off. Doing that only for this will create confusion. Also, I personally think this type of changing content whether switch is on/off creates more confusion for user.

I agree. I honestly don't know what I was thinking.

Regardless, the description text should be moved to under the button it corresponds to.

soumyamahunt:
> Regardless, the description text should be moved to under the button it corresponds to.

Currently the new toggle switch doesn't contain any description as the content itself pretty self explanatory, as for the "always open new window" option the description for it is below it.

UmbralReaper:
> Currently the new toggle switch doesn't contain any description as the content itself pretty self explanatory, as for the "always open new window" option the description for it is below it.

Turns out I was reading it the wrong way round 😅 .

luke-uppx:
Why has this functionality not been applied yet?

philippehawi:
Any updates on this? Will it be added to a release soon?

PR #602, opened 2020-06-25
Implemented settings from previous installation initializes settings for new installation

soumyamahunt:
If Notepads is installed under a Microsoft account, the settings is synced with this account. For a new installation under this account the default settings value will be from these synced settings

## PR Type

Feature

### DIFF
--- a/src/Notepads/App.xaml.cs
+++ b/src/Notepads/App.xaml.cs
@@ -85,6 +85,7 @@
                 Window.Current.Content = rootFrame;
                 rootFrameCreated = true;
 
+                ApplicationSettingsStore.Initialize();
                 ThemeSettingsService.Initialize();
                 AppSettingsService.Initialize();
             }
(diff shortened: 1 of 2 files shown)


JasonStein:
I thought this roaming thing is going to deprecated, can you confirm?

soumyamahunt:
> I thought this roaming thing is going to deprecated, can you confirm?

[This](https://docs.microsoft.com/en-us/windows/deployment/planning/windows-10-deprecated-features) article says it is deprecated but the functionality isn't removed. Should I close this or will you implement it now and remove this when roaming is removed??

JasonStein:
> > I thought this roaming thing is going to deprecated, can you confirm?
> 
> [This](https://docs.microsoft.com/en-us/windows/deployment/planning/windows-10-deprecated-features) article says it is deprecated but the functionality isn't removed. Should I close this or will you implement it now and remove this when roaming is removed??

Let's not implement it since it is scheduled to be discontinued. It will cause confusion in the future.

PR #603, opened 2020-06-26
Implemented settings and recently opened file list syncing across all instances

soumyamahunt:
Changing settings in one window now affects all other windows. Also recently opened files are synced across all instances.

## PR Type

Feature


soumyamahunt:
@JasonStein this is implemented without any background task, so please have a look.

PR #605, opened 2020-06-28
Made some UI refinements

soumyamahunt:
Added separator between hamburger button and tabs. Compact overlay now opens with maximum height and width.

## PR Type

Feature

## Other information
cmxiTInXru
KKP2S8BxrL

### DIFF
--- a/src/Notepads.Controls/SetsView/SetsView.cs
+++ b/src/Notepads.Controls/SetsView/SetsView.cs
@@ -223,7 +223,17 @@ namespace Notepads.Controls
 
         private void UpdateSetSeparators()
         {
-            for (int i = 0; i < Items?.Count; i++)
+            int i = 0;
+            if (SelectedIndex != i)
+            {
+                (ContainerFromIndex(i) as SetsViewItem)?.ShowLeftSideSeparator();
+            }
+            else
+            {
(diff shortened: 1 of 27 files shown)


JasonStein:
I don't think that search apply button is necessary. It is kind of counter intuitive. Can you revert the changes for that part? Also remove the tooltip string since it won't be needed. Rest are looking good.

soumyamahunt:
> I don't think that search apply button is necessary. It is kind of counter intuitive. Can you revert the changes for that part? Also remove the tooltip string since it won't be needed. Rest are looking good.

Reverted those changes.

src/Notepads.Controls/SetsView/SetsView.cs
```diff
@@ -223,7 +223,17 @@ private void SetsView_SelectionChanged(object sender, SelectionChangedEventArgs
 
         private void UpdateSetSeparators()
         {
-            for (int i = 0; i < Items?.Count; i++)
+            int i = 0;
+            if (SelectedIndex != i)
```
JasonStein, line 227: use != 0 here. "!= i" might have some side effect in the future. Just move int i = 0 into the for loop and use "0" for the first if-else block.

soumyamahunt: Made changes.

PR #612, opened 2020-07-15
Implemented accepting environment variables as commandline arguments

soumyamahunt:
Implemented accepting environment variables like %userpath%, %temp%, %appdata%, %windir% etc. as commandline arguments.

## PR Type

Feature

### DIFF
--- a/src/Notepads/Utilities/FileSystemUtility.cs
+++ b/src/Notepads/Utilities/FileSystemUtility.cs
@@ -118,6 +118,7 @@
 
             try
             {
+                args = ReplaceEnvironmentVariables(args);
                 path = GetAbsolutePathFromCommandLine(dir, args, App.ApplicationName);
             }
             catch (Exception ex)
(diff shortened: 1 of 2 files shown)


src/Notepads/Utilities/FileSystemUtility.cs
```diff
@@ -135,6 +136,43 @@ public static async Task<StorageFile> OpenFileFromCommandLine(string dir, string
...
+            {
+                args = args.Replace("%localappdata%",
+                    UserDataPaths.GetDefault().LocalAppData,
+                    StringComparison.OrdinalIgnoreCase);
+            }
+
+            if (args.Contains("%temp%", StringComparison.OrdinalIgnoreCase))
+            {
+                args = args.Replace("%temp%",
+                    (string)Microsoft.Win32.Registry.GetValue(@"HKEY_CURRENT_USER\Environment",
```
JasonStein, line 158: Do we need extra permission to access registry?

soumyamahunt: No, reading registry doesn't require any permission.

PR #622, opened 2020-07-19
Copy/Cut whole line when there is nothing selected with cursor

gmaza:
If there is no text selected, Copy and Cut will work for the current line that the cursor is on.
This feature works for a context menu and also for shortcuts (ctrl+c, ctrl+x).

## PR Type
Feature

## Other information
image


JasonStein:
@soumyamahunt There should be an easier way to implement this using RichEdit paragraph property, can you confirm?

soumyamahunt:
> @soumyamahunt There should be an easier way to implement this using RichEdit paragraph property, can you confirm?

Haven't tested this, but this should work:
```
Document.Selection.Expand(TextRangeUnit.Paragraph);
Document.Selection.Copy()/Document.Selection.Cut();
```
@gmaza Can you confirm??

gmaza:
Okay, I'll test it.

gmaza:
Thank you, guys. it's a more elegant way to implement this feature, so I updated and tested it.

src/Notepads/Controls/TextEditor/TextEditor.xaml.cs
```diff
@@ -783,17 +783,31 @@ public void CutSelectedTextToWindowsClipboard(TextControlCuttingToClipboardEvent
...
+
+                if (!isTextSelected)
+                {
+                    TextEditorCore.Document.Selection.Expand(TextRangeUnit.Paragraph);
+                }
+
                 var text = LineEndingUtility.ApplyLineEnding(TextEditorCore.Document.Selection.Text, GetLineEnding());
                 dataPackage.SetText(text);
+
+                if (clearLineSelection && !isTextSelected)
```
JasonStein, line 806: move this logic to CutSelectedTextToWindowsClipboard and remove the "bool clearLineSelection" parameter.

gmaza: okay, but if I move this logic to CutSelectedTextToWindowsClipboar then there will be code duplication, because "Copy" also needs it.

JasonStein: CopyTextToWindowsClipboardInternal will do the copy, isn't it?

gmaza:
Of course, it will do the copy.
I'll explain better:
I think it's better for UX not to leave text selected when you copy whole line (Like other editors), that's because I need this logic in CopyTextToWindowsClipboardInternal.
There are 3 possible options:
1) copy this logic in 
CutSelectedTextToWindowsClipboar and it will be duplicated but code will be wet principle style.

2) move this logic into CutSelectedTextToWindowsClipboar, code will be more clear, but when you copy whole line (without selection), it will be selected after copy.

3) keep everything as they are now. 

If you have another option it will be interesting for me

JasonStein:
> Of course, it will do the copy.
> I'll explain better:
> I think it's better for UX not to leave text selected when you copy whole line (Like other editors), that's because I need this logic in CopyTextToWindowsClipboardInternal.
> There are 3 possible options:
> 
> 1. copy this logic in
>    CutSelectedTextToWindowsClipboar and it will be duplicated but code will be wet principle style.
> 2. move this logic into CutSelectedTextToWindowsClipboar, code will be more clear, but when you copy whole line (without selection), it will be selected after copy.
> 3. keep everything as they are now.
> 
> If you have another option it will be interesting for me

I see, keep as it is for now.

PR #632, opened 2020-07-29
Update issue templates

MarcAnt01:
Now that we have a bot for labels, titles of issues are no longer needed

## PR Type

## Other information->
Not needed


JasonStein:
Why?

MarcAnt01:
There are labels to differentiate them per type, I'd rather have only the tile of the issue there

JasonStein:
> There are labels to differentiate them per type, I'd rather have only the tile of the issue there

I don't think that's too much of a concern here.... I am keeping it as it is for now.

PR #636, opened 2020-08-02
Implement modifying files with readonly attribute

soumyamahunt:
This pr aims to implement  modifying files with readonly attribute.

## PR Type

Feature

## Other information
While Windows Notepad doesn't allow modifying files with readonly attribute, Notepad++ allows editing them once this flag is removed.

### DIFF
--- a/src/Notepads/Controls/TextEditor/ITextEditor.cs
+++ b/src/Notepads/Controls/TextEditor/ITextEditor.cs
@@ -46,6 +46,8 @@
 
         string EditingFilePath { get; }
 
+        bool IsReadOnly { get; }
+
         StorageFile EditingFile { get; }
 
         bool IsModified { get; }
--- a/src/Notepads/Controls/TextEditor/TextEditor.xaml.cs
+++ b/src/Notepads/Controls/TextEditor/TextEditor.xaml.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Runtime.CompilerServices;
     using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
@@ -73,6 +74,8 @@
 
         public string EditingFilePath { get; private set; }
 
+        public bool IsReadOnly { get; private set; }
+
         private StorageFile _editingFile;
 
         public StorageFile EditingFile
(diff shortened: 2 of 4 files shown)


soumyamahunt:
@JasonStein Notepad++ doesn't allow editing readonly files by default, but after removing readonly flag it allows editing and user than can save the file. Should we implement this behavior?? Or, should we prompt user when trying to save a readonly file and if user accepts remove the readonly attribute??

soumyamahunt:
@JasonStein this is the behavior I have come up with for editing read-only vs normal files:

| <h3 align="center"> Normal </h3> |   <h3 align="center"> Read-Only </h3> |
| :--------------------------------- | :-------------------- |
| ApplicationFrameHost_a9vdeKJHlh | ApplicationFrameHost_7Ho0hOMPrl |
| ApplicationFrameHost_L8tJ5ic4tW | ApplicationFrameHost_U98Fk7JAP8 |
| ApplicationFrameHost_slyQ40nEgE | ApplicationFrameHost_EJPpQZGQTA |
| ApplicationFrameHost_rwcf7wkwAS | ApplicationFrameHost_SB2YD43fe4 |

JasonStein:
Two cents:

1. Display a lock icon right after the path indicator so it is more intuitive.
2. Whenever user changes any text or add/delete any text, show normal notification saying that "File is read-only!". Let user change the attribute of the file so we do not need to implement anything.

Btw, how could you import those win32 dlls? Will they work in production? What about future compatibility with Windows 10x?

soumyamahunt:
> 1. Display a lock icon right after the path indicator so it is more intuitive.

Will do that.
> 2. Whenever user changes any text or add/delete any text, show normal notification saying that "File is read-only!". Let user change the attribute of the file so we do not need to implement anything.

Wouldn't displaying notification be too intrusive?? Currently I am showing the modified icon in the tab in grey/disabled color when file is read only. 

> Btw, how could you import those win32 dlls? Will they work in production? What about future compatibility with Windows 10x?

[..FromApp](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/mt846626(v=vs.85)) methods are made with uwp security model so they will work in production and there won't be any future incompatibility.

JasonStein:
> > 1. Display a lock icon right after the path indicator so it is more intuitive.
> 
> Will do that.
> 
> > 1. Whenever user changes any text or add/delete any text, show normal notification saying that "File is read-only!". Let user change the attribute of the file so we do not need to implement anything.
> 
> Wouldn't displaying notification be too intrusive?? Currently I am showing the modified icon in the tab in grey/disabled color when file is read only.
> 
> > Btw, how could you import those win32 dlls? Will they work in production? What about future compatibility with Windows 10x?
> 
> [..FromApp](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/mt846626(v=vs.85)) methods are made with uwp security model so they will work in production and there won't be any future incompatibility.

What about the file that user dragged into notepads? (They are read-only to notepads but they do not have read-only attribute)

soumyamahunt:
> What about the file that user dragged into notepads? (They are read-only to notepads but they do not have read-only attribute)

They would work like normal files. None of the `StorageFile` annoyances apply to `..FromApp` methods.

soumyamahunt:
@JasonStein I have modified the behavior as you said:

| <h3 align="center"> Normal </h3> |   <h3 align="center"> Read-Only </h3> |
| :--------------------------------- | :-------------------- |
| ApplicationFrameHost_a9vdeKJHlh | ApplicationFrameHost_iVJ6G9BoYg |
| ApplicationFrameHost_L8tJ5ic4tW | ApplicationFrameHost_c5d036RQms |
| ApplicationFrameHost_slyQ40nEgE | ApplicationFrameHost_6WSOc2ApVE |
| ApplicationFrameHost_rwcf7wkwAS | ApplicationFrameHost_R8L6V8oJnq |

PR #639, opened 2020-08-03
added dark icons on light start tiles

mah-qurashy:


JasonStein:
image
image
The task bar icon and menu bar icons are not in sync, do you prefer white under light, dark under dark or the other way around?

mah-qurashy:
> The task bar icon and menu bar icons are not in sync, do you prefer white under light, dark under dark or the other way around?

honestly i think that light on light / dark on dark looks better
but the better decision is to stick with what we currently have to keep consistency with default windows apps

JasonStein:
> > The task bar icon and menu bar icons are not in sync, do you prefer white under light, dark under dark or the other way around?
> 
> honestly i think that light on light / dark on dark looks better
> but the better decision is to stick with what we currently have to keep consistency with default windows apps

I agree with you. Let me think about it.

PR #663, opened 2020-08-26
Fix for detecting shadow windows

soumyamahunt:
Made some changes to detect shadow windows properly in release mode.

## PR Type

Bugfix


soumyamahunt:
@JasonStein this should solve detecting shadow windows even in release mode.

Issue #667, opened 2020-08-27
[Bug] Incomplete functionality when highlighting misspelled words


XamDR:
**Describe the bug**
When the option to highlight spell check errors is enabled, it only highlights the misspelled word, but it doesn't give a list of suggestions to fix the error. As an user I find this kind of behavior counter-intuitive because I expect to have the whole highlight and suggest experience, similar to how other programs do like Word or Edge Chromium. Currently, it feels like the mentioned functionality is incomplete.

**To Reproduce**
1. Toggle on the option to highlight misspelled words.

**Expected behavior**
The misspelled words should be underline with the classical red wavy underline and right clicking on it should display a list of suggestions, the option to add to the dictionary or to ignore the word, in the text editor's context menu.

**Screenshots**
For example, while I was redacting this issue, I misspelled a word and this is what Edge did:

![image](https://user-images.githubusercontent.com/35734934/91469225-bb939980-e858-11ea-8a9c-f19bc3f693d4.png)

In contrast, this is what Notepads does:

![image](https://user-images.githubusercontent.com/35734934/91470044-d1558e80-e859-11ea-9ca5-647a94c4330a.png)

**Desktop (please complete the following information):**
 - OS: Windows 10 1909 18363.1016
 - Version: v1.3.9.0

**Additional context**
I've checked the code of the file `TextEditorContextFlyout.cs` and I found a way to fix the issue described above using the `RichEditBox`'s `ProofingMenuFlyout` property. And if it's necessary, I would like to summit a pull request.

soumyamahunt:
> And if it's necessary, I would like to summit a pull request.

Feel free to do that.

XamDR:
> > And if it's necessary, I would like to summit a pull request.
> 
> Feel free to do that.

Ok thanks. I will do ASAP.

PR #669, opened 2020-08-28
Added list of suggestions for misspelled words in the text editor context menu

XamDR:
This PR intents to fix the bug reported [here](https://github.com/JasonStein/Notepads/issues/667).

## PR Type
Feature

## Other information
image

image

image

The proofing menu flyout items are displayed in the classical way a la Word 2016 or Edge Chromium, where these are above the standard items (cut, copy, paste, etc.). Unfortunately, I was unable to mimic the standard proofing flyout of the RichEditBox because there are two issues:

- The `Items` in the `MenuFlyoutSubItem` can't be updated at runtime once they are added, as described in this [issue](https://github.com/microsoft/microsoft-ui-xaml/issues/1652).
- There is no way to programmatically open a `MenuFlyoutSubItem` because it doesn't have a `IsOpened` property or something similar.

### DIFF
--- a/src/Notepads.Controls/Notepads.Controls.csproj
+++ b/src/Notepads.Controls/Notepads.Controls.csproj
@@ -11,7 +11,7 @@
     <AssemblyName>Notepads.Controls</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">10.0.18362.0</TargetPlatformVersion>
+    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">10.0.19041.0</TargetPlatformVersion>
     <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
     <MinimumVisualStudioVersion>14</MinimumVisualStudioVersion>
     <FileAlignment>512</FileAlignment>
--- a/src/Notepads.Controls/Notepads.Controls.csproj
+++ b/src/Notepads.Controls/Notepads.Controls.csproj
@@ -11,7 +11,7 @@
     <AssemblyName>Notepads.Controls</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">10.0.19041.0</TargetPlatformVersion>
+    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">10.0.18362.0</TargetPlatformVersion>
     <TargetPlatformMinVersion>10.0.17763.0</TargetPlatformMinVersion>
     <MinimumVisualStudioVersion>14</MinimumVisualStudioVersion>
     <FileAlignment>512</FileAlignment>
--- a/src/Notepads/Controls/TextEditor/TextEditorContextFlyout.cs
+++ b/src/Notepads/Controls/TextEditor/TextEditorContextFlyout.cs
@@ -24,6 +24,9 @@
         private MenuFlyoutItem _previewToggle;
         private MenuFlyoutItem _share;
 
+        private MenuFlyout _proofingFlyout;
+        private readonly MenuFlyoutSeparator _separator0 = new MenuFlyoutSeparator();
+
         private readonly ITextEditor _textEditor;
         private readonly TextEditorCore _textEditorCore;
 
@@ -47,15 +50,44 @@
             Items.Add(Share);
 
             Opening += TextEditorContextFlyout_Opening;
+            Closed += TextEditorContextFlyout_Closed;
         }
 
         public void Dispose()
         {
             Opening -= TextEditorContextFlyout_Opening;
+            Closed -= TextEditorContextFlyout_Closed;
         }
 
         private void TextEditorContextFlyout_Opening(object sender, object e)
         {
+            if (_textEditorCore.IsSpellCheckEnabled)
+            {
+                _proofingFlyout = _textEditorCore.ProofingMenuFlyout as MenuFlyout;
+
+                if (_proofingFlyout.Items.Count > 0)
+                {
+                    foreach (var item in _proofingFlyout.Items)
+                    {
+                        Items.Insert(_proofingFlyout.Items.IndexOf(item), item);
+                    }
+
+                    if (!Items.Contains(_separator0))
+                    {
+                        Items.Insert(_proofingFlyout.Items.Count, _separator0);
+                    }
+
+                    if (_separator0.Visibility == Visibility.Collapsed)
+                    {
+                        _separator0.Visibility = Visibility.Visible;
+                    }
+                }
+                else
+                {
+                    _separator0.Visibility = Visibility.Collapsed;
+                }
+            }
+
             if (_textEditorCore.Document.Selection.Type == SelectionType.InsertionPoint ||
                 _textEditorCore.Document.Selection.Type == SelectionType.None)
             {
@@ -76,6 +108,17 @@
             if (App.IsGameBarWidget)
             {
                 Share.Visibility = Visibility.Collapsed;
+            }
+        }
+
+        private void TextEditorContextFlyout_Closed(object sender, object e)
+        {
+            if (_proofingFlyout != null)
+            {
+                foreach (var item in _proofingFlyout.Items)
+                {
+                    Items.Remove(item);
+                }
             }
         }
 
--- a/src/Notepads/Controls/TextEditor/TextEditorContextFlyout.cs
+++ b/src/Notepads/Controls/TextEditor/TextEditorContextFlyout.cs
@@ -25,7 +25,7 @@
         private MenuFlyoutItem _share;
 
         private MenuFlyout _proofingFlyout;
-        private readonly MenuFlyoutSeparator _separator0 = new MenuFlyoutSeparator();
+        private readonly MenuFlyoutSeparator _proofingSeparator = new MenuFlyoutSeparator();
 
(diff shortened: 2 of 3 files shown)


src/Notepads.Controls/Notepads.Controls.csproj
```diff
@@ -11,7 +11,7 @@
     <AssemblyName>Notepads.Controls</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">10.0.19041.0</TargetPlatformVersion>
+    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">10.0.18362.0</TargetPlatformVersion>
```
soumyamahunt, line 14: Is there any specific reason you are decreasing target version??

XamDR: It is because I don't have the latest SDK. Should I install it?

soumyamahunt: Yes, install the latest sdk, Notepads depends on this to provide a some feature that aren't available in previous versions.

XamDR: Ok. I'll install it. Once I have done it, do I have to submit a new PR, right?

soumyamahunt:
> The proofing menu flyout items are displayed in the classical way a la Word 2016 or Edge Chromium, where these are above the standard items (cut, copy, paste, etc.). Unfortunately, I was unable to mimic the standard proofing flyout of the RichEditBox because there are two issues:
> 
> * The `Items` in the `MenuFlyoutSubItem` can't be updated at runtime once they are added, as described in this [issue](https://github.com/microsoft/microsoft-ui-xaml/issues/1652).
> * There is no way to programmatically open a `MenuFlyoutSubItem` because it doesn't have a `IsOpened` property or something similar.

Instead of trying to add items to `MenuFlyoutSubItem` at runtime, you can add the whole `MenuFlyoutSubItem` at runtime with all the updated items. You have to remove the `MenuFlyoutSubItem` each time context menu is invoked too.

src/Notepads.Controls/Notepads.Controls.csproj
soumyamahunt: You don't have to submit new pr, just revert the target platform changes you have made and push a new commit.

XamDR: Ok. I just finished to install it. Now, I will push a new commit. Thanks.

XamDR:
> > The proofing menu flyout items are displayed in the classical way a la Word 2016 or Edge Chromium, where these are above the standard items (cut, copy, paste, etc.). Unfortunately, I was unable to mimic the standard proofing flyout of the RichEditBox because there are two issues:
> > 
> > * The `Items` in the `MenuFlyoutSubItem` can't be updated at runtime once they are added, as described in this [issue](https://github.com/microsoft/microsoft-ui-xaml/issues/1652).
> > * There is no way to programmatically open a `MenuFlyoutSubItem` because it doesn't have a `IsOpened` property or something similar.
> 
> Instead of trying to add items to `MenuFlyoutSubItem` at runtime, you can add the whole `MenuFlyoutSubItem` at runtime with all the updated items. You have to remove the `MenuFlyoutSubItem` each time context menu is invoked too. We do this approach for showing submenu of recent files in Notepads.

I thought about that approach, but the problem is about UX. Because the `MenuFlyoutSubItem` doesn't have a `IsOpened` property, the user has to click to open the flyout and then click again to choose a suggestion. I couldn't find a way to display the list of suggestions immediately like Word 2019 or the built-in RichEditBox's proofing menu flyout. Anyways, I will give a try to your suggestion and see if there is a chance to have the behavior mentioned before.

src/Notepads/Controls/TextEditor/TextEditorContextFlyout.cs
```diff
@@ -24,6 +24,9 @@ public class TextEditorContextFlyout : MenuFlyout
         private MenuFlyoutItem _previewToggle;
         private MenuFlyoutItem _share;
 
+        private MenuFlyout _proofingFlyout;
+        private readonly MenuFlyoutSeparator _separator0 = new MenuFlyoutSeparator();
```
JasonStein, line 28: Rename it to _proofingSeparator

XamDR: Yeah. That's definitely a better name.

XamDR:
> > > The proofing menu flyout items are displayed in the classical way a la Word 2016 or Edge Chromium, where these are above the standard items (cut, copy, paste, etc.). Unfortunately, I was unable to mimic the standard proofing flyout of the RichEditBox because there are two issues:
> > > 
> > > * The `Items` in the `MenuFlyoutSubItem` can't be updated at runtime once they are added, as described in this [issue](https://github.com/microsoft/microsoft-ui-xaml/issues/1652).
> > > * There is no way to programmatically open a `MenuFlyoutSubItem` because it doesn't have a `IsOpened` property or something similar.
> > 
> > 
> > Instead of trying to add items to `MenuFlyoutSubItem` at runtime, you can add the whole `MenuFlyoutSubItem` at runtime with all the updated items. You have to remove the `MenuFlyoutSubItem` each time context menu is invoked too. We do this approach for showing submenu of recent files in Notepads.
> 
> I thought about that approach, but the problem is about UX. Because the `MenuFlyoutSubItem` doesn't have a `IsOpened` property, the user has to click to open the flyout and then click again to choose a suggestion. I couldn't find a way to display the list of suggestions immediately like Word 2019 or the built-in RichEditBox's proofing menu flyout. Anyways, I will give a try to your suggestion and see if there is a chance to have the behavior mentioned before.

I tried the approach of displaying the `RichEditbox`'s `ProofingMenuFlyout` as a `MenuFlyoutSubItem`, as suggested, and it doesn't work because the application crashes when a suggestion is chosen.

```
var proofingFlyoutSubItem = new MenuFlyoutSubItem { Text = "Proofing" };

foreach (var item in _proofingFlyout.Items)
{
    proofingFlyoutSubItem.Items.Add(item);
}

Items.Insert(0, proofingFlyoutSubItem);
```

image

After clicking on the word _funciona_ the application crashed immediately without any message in the Visual Studio result window. I then changed the debug option to Mixed (native and managed), and the app crashed with the message: "Unhandled exception at 0x00007FF90FA6EF06 (Windows.UI.Xaml.dll) in Notepads.exe: 0xC000027B: An internal application exception occurred (parameters: 0x000001CD33620F10, 0x0000000000000003)."

soumyamahunt:
> I tried the approach of displaying the `RichEditbox`'s `ProofingMenuFlyout` as a `MenuFlyoutSubItem`, as suggested, and it doesn't work because the application crashes when a suggestion is chosen.
>
> After clicking on the word _funciona_ the application crashed immediately without any message in the Visual Studio result window. I then changed the debug option to Mixed (native and managed), and the app crashed with the message: "Unhandled exception at 0x00007FF90FA6EF06 (Windows.UI.Xaml.dll) in Notepads.exe: 0xC000027B: An internal application exception occurred (parameters: 0x000001CD33620F10, 0x0000000000000003)."

Strange! It didn't happen when adding items directly to flyout?? Also, can you try launching the app in normal environment and see the words are changed or not??

XamDR:
> > I tried the approach of displaying the `RichEditbox`'s `ProofingMenuFlyout` as a `MenuFlyoutSubItem`, as suggested, and it doesn't work because the application crashes when a suggestion is chosen.
> > After clicking on the word _funciona_ the application crashed immediately without any message in the Visual Studio result window. I then changed the debug option to Mixed (native and managed), and the app crashed with the message: "Unhandled exception at 0x00007FF90FA6EF06 (Windows.UI.Xaml.dll) in Notepads.exe: 0xC000027B: An internal application exception occurred (parameters: 0x000001CD33620F10, 0x0000000000000003)."
> 
> Strange! It didn't happen when adding items directly to flyout?? Also, can you try launching the app in normal environment and see the words are changed or not??

When I added the items to the context flyout it works fine. The issue happened when I tried to add them to the `MenuFlyoutSubItem`. I don't know if it's related to the fact that the `MenuFlyoutSubItem` is instantiated every time. 

I don't understand exactly what do you mean with "launching the app in normal environment". Could you clarify, please? 
I tried both in debug and release mode, and the app worked fine.

soumyamahunt:
> I don't understand exactly what do you mean with "launching the app in normal environment". Could you clarify, please?
> I tried both in debug and release mode, and the app worked fine.

Try launching the app from start menu instead of VS' debugging environment, see if you can replace the mis-spelled words with this approach:
> Instead of trying to add items to MenuFlyoutSubItem at runtime, you can add the whole MenuFlyoutSubItem at runtime with all the updated items. You have to remove the MenuFlyoutSubItem each time context menu is invoked too. We do this approach for showing submenu of recent files in Notepads.

XamDR:
> > I don't understand exactly what do you mean with "launching the app in normal environment". Could you clarify, please?
> > I tried both in debug and release mode, and the app worked fine.
> 
> Try launching the app from start menu instead of VS' debugging environment, see if you can replace the mis-spelled words with this approach:
> 
I just tried and it doesn't work either. The app crashes again and it closes after a few seconds.

JasonStein:
I think current approach is good enough. I do wish we can put it into a sub menu but them it is less accessible overall which is not ideal.

@XamDR @soumyamahunt I am fine with current approach and implementation.

XamDR:
> I think current approach is good enough. I do wish we can put it into a sub menu but then it becomes less accessible overall, which is not ideal.
> 
> @XamDR @soumyamahunt I am fine with current approach and implementation.

That's exactly the issue with putting the items in a sub menu, because there is no programmatic way to display the `MenuFlyoutSubItem` opened. Then, the user has to 'wait' because there is an small delay (an issue related to this has been reported [here](https://github.com/microsoft/microsoft-ui-xaml/issues/939)).

PR #676, opened 2020-09-04
Update Spanish translation

XamDR:
Added some missing Spanish translations.

## PR Type

 Translation


XamDR:
Hi. I would like to know if there are any updates about this.

Issue #662, opened 2020-08-23
[Bug] shadow window


AlessandroVeller:
**Describe the bug**
Every time I open the app a message pops up saying "this is a shadow window of notepads. Session snapshot and settings are disabled." Am I doing something wrong or is this a bug?

KristofKekesi:
I have this bug too.

JasonStein:
This will be fixed in v1.4.0.0

PkW4v3:
I have version 1.4.2.0 but i still have this bug. Can you help me? @JasonStein

kushaagr:
Issue persists in version 1.4.2.0
![2020 12 24](https://user-images.githubusercontent.com/68564934/103039129-15ea0000-4596-11eb-895a-9593e642695e.png)

soumyamahunt:
@PkW4v3 @PkW4v3 can you provide a video recording of launching notepads with the task manager window included. Also expand the "Notepads" node in task manager and show the changes for that node in the video.

706B:
I am also facing this issue. I already have notepads on my desktop which works fine. I just installed it on my laptop today and I am getting this message every time I open the app. Just after the install, it worked fine but after that (from the second time I opened the app), I have been getting this message. the setting is also not accessible now in the app.

one noticeable difference between my laptop and desktop is that my desktop is on window10 2004 version and my laptop in on 20H2 (latest ver.)

![image](https://user-images.githubusercontent.com/14940852/103410704-68ae5380-4b64-11eb-87c0-3033ed9194a6.png)

GalaxytedYT:
I first had it from MS Store, but then I re-installed it from GitHub and it works now. (no shadow bug)

PR #742, opened 2021-01-03
Fix for primary instance being detected as shadow instance.

soumyamahunt:
Tries fixing #662 permanently.

## PR Type

Bugfix

## Other information
Some users are still having issue where primary instance is miscategorized as shadow instance. This PR aims to fix that.

### DIFF
--- a/src/Notepads/App.xaml.cs
+++ b/src/Notepads/App.xaml.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Collections.Generic;
     using System.Linq;
+    using System.Threading;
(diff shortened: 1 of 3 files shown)


JasonStein:
@soumyamahunt What if notepads crashed or user killed notepads instance from taskmgr?

soumyamahunt:
> @soumyamahunt What if notepads crashed or user killed notepads instance from taskmgr?

The detection would still work regardless Notepads is closed by user, or crashed/forcefully killed by user.

PR #752, opened 2021-01-13
Provided entry to create new text file in shell context menu

soumyamahunt:
Since Windows Notepad can be uninstalled, uninstalling Notepad removes the entry to create new text file in shell context menu. This PR enables Notepads to provide this capability to users even if Windows Notepad is uninstalled

## PR Type

Feature

## Other information
If both Windows Notepad and Notepads are installed, there is only one context menu entry displayed(no duplicate entries). The context menu in file explorer may not be updated as soon as app is installed/uninstalled and explorer may require a restart.

explorer_7TikXAnJx3


JasonStein:
“If both Windows Notepad and Notepads are installed, there is only one context menu entry displayed(no duplicate entries”

What if the translation of "Text Document" is different from the one comes in Windows Notepad? I guess it does not matter right?

soumyamahunt:
> What if the translation of "Text Document" is different from the one comes in Windows Notepad? I guess it does not matter right?

Yes, it doesn't matter.

PR #755, opened 2021-01-15
ImplementedWin32 component to redirect activation to UWP app

soumyamahunt:
This PR implements Win32 component which redirects file activation to Notepads. User can use this .exe file to open unsupported files with Notepads.

## PR Type

Feature

## Other information
While for Windows 10 2004 and later version user can open any file types with Notepads, users in earlier version of Windows 10 can use this desktop component to open unsupported file types.

Also, UWP restricted file types like batch file(.bat, .cmd etc.), VB Script files(.vbe, .vbs etc.) and .reg files show "Edit" shell menu entries which opens these file types in Windows Notepad and edit them. User can use desktop component either by modifying registry or copying the desktop component to System32 directory to open these file types with Notepads instead.

Ylf7SHUFgQ


soumyamahunt:
@JasonStein  Since Notepads doesn't need to launch this desktop component in any way, it is included as a file for user and no `runFullTrust` capability is required. We can also provide option in Notepads interface for user to copy the executable path or provide the executable to place anywhere the user wants. Let me know what you think about this.

PR #772, opened 2021-02-06
Added Arabic translation [ar-YE]

HasanAlyazidi:
## PR Type

Translation

### Main:
Main

### Settings:
Settings

## Other information
There is a bug that appear only for RTL languages and I will create an issue for that.


src/Notepads/Notepads.csproj
```diff
@@ -449,6 +449,15 @@
   <ItemGroup>
     <None Include="Package.targets" />
   </ItemGroup>
+  <ItemGroup>
```
JasonStein, line 452: You can remove these lines since it will be covered by <PRIResource Include="Strings\**\*.resw" />

JasonStein (changes requested):

src/Notepads/Notepads.csproj
HasanAlyazidi: Only L449 to L451?

soumyamahunt: @HasanAlyazidi only remove the lines you have added L452 to L460.

HasanAlyazidi: @soumyamahunt @JasonStein done.

PR #777, opened 2021-02-13
Use MSBuild tasks instead of Powershell script to generate manifest

soumyamahunt:
Fixes some cases where Powershell script doesn't return the generated manifest file path.

## PR Type

Bugfix


JasonStein:
@soumyamahunt Can you help resolve the conflicts?

soumyamahunt:
@JasonStein RTM.

PR #781, opened 2021-02-18
Added feature to compare files

oleitao:
## PR Type
Added an option to allow compare two files lines changes

## Other information
In this implementation i added an option on menu allowing the user to select two files load them into the app and view all lines changed between them. 
I added translations for all languages except 'or-IN'. However, some translations could not be correct or have some wrong terms.
On a future implementation i intend to add an option to select three files at the same time instead of only two.


soumyamahunt:
@oleitao do you know all the languages you have provided translation for?? If not revert those strings to English so that translators know which string to translate.

oleitao:
> 
> 
> @oleitao do you know all the languages you have provided translation for?? If not revert those strings to English so that translators know which string to translate.

Hi, @soumyamahunt i used Google Translator to translante. I already reverted all string to English just leaving (pt-BR and en-US)

oleitao:
> @oleitao do you know all the languages you have provided translation for?? If not revert those strings to English so that translators know which string to translate.

@soumyamahunt  i've an CodeFactor pipeline error witch is related wuth method complexity. However, i think it's better don't change. What's your opinion?

JasonStein:
I thought of this a while back but decided not to do it since this feature alone deserves an app for it and there are plenty of existing apps that can do it and do it better than DiffPlex.

I also thought about adding this feature to allow user to compare two existing tabs opened in current session but it will make the logic a little bit messy since it is better to have another window showing the result.

I would say this is good to have but not convincing enough to be added as part of Notepads. The current diff view is only intended to show the changes user made to their txt file which is very relevant to the app logic but comparing two files is not in scope.

PR #786, opened 2021-02-21
Making session snapshot full proof

soumyamahunt:
1. Implemented [transacted write model](https://docs.microsoft.com/en-us/windows/uwp/files/best-practices-for-writing-to-files#the-transactional-model) for creating session snapshot files (both backup and metadata file). 
2. Reduced number of IO operations to improve speed during load.
3. Implemented recovering tabs from orphaned snapshot files.

## PR Type

Feature

### DIFF
--- a/src/Notepads/Core/SessionManager.cs
+++ b/src/Notepads/Core/SessionManager.cs
@@ -453,7 +453,7 @@
         {
             try
             {
-                await FileSystemUtility.WriteToFile(LineEndingUtility.ApplyLineEnding(text, lineEnding), encoding, file);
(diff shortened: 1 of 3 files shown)


soumyamahunt:
@JasonStein I think notepads should use database like [Realm](https://github.com/realm/realm-dotnet) instead of using a json file to store metadata. This would make storing metadata safer. Also, we can use [.~TMP](https://docs.microsoft.com/en-us/windows/uwp/files/best-practices-for-writing-to-files#tmp-files) files remaining from failed write transaction to recover more data.

PR #795, opened 2021-03-02
Added pt-PT language translations

oleitao:
## PR Type
Added pt-PT language translations.


src/Notepads/Notepads.csproj
```diff
@@ -443,6 +443,15 @@
   <ItemGroup>
     <PRIResource Include="Strings\**\*.resw" />
   </ItemGroup>
+  <ItemGroup>
```
JasonStein, line 446: You don't need to add these since it is already covered by "<PRIResource Include="Strings\**\*.resw" />" You can remove all new lines added in this file.

JasonStein (changes requested):

src/Notepads/Notepads.csproj
oleitao: Hi, I already remove that lines, verify my last commit

Issue #761, opened 2021-01-23
[Bug] this is a shadow screen on notepads. Session settings and restore are disabled


noneoriente:
every time I open notepads I get a notification

this is a shadow screen on notepads. Session settings and restore are disabled

noneoriente:
Doesn't anyone answer? Well, if this Git has little movement, uninstalling the app is not worth wasting time.

Thanks also.

soumyamahunt:
The next update should fix this issue, meanwhile you can try ending all instance of Notepads from task manager(even suspended one) and try launching the app again.

noneoriente:
occurs when activating the option (open in new window)

JasonStein:
The next update will address this issue.

PR #804, opened 2021-03-14
Fix issue which makes a prelaunched instance primary instance

soumyamahunt:
As pointed out by @noneoriente in #761, this issue prevents pre-launched instances to make themselves primary instance and only visible instances can be primary instance.

## PR Type

Bugfix

## Other information
 This issue prevented users from opening a primary instance app window when user has enabled `Always open in new window` settings. In this case Windows pre-launches Notepads instance which registers itself as primary. Subsequent activation by user will always create new window without bringing the primary instance to front. [Documentation](https://docs.microsoft.com/en-us/windows/uwp/launch-resume/handle-app-prelaunch).


soumyamahunt:
@JasonStein also the `Always open in new window` option can't take advantage of pre-launching feature as there will always be new instance created even if there is a background pre-launched instance available. We should detect pre-launched instances and activate them first before creating new instances.

JasonStein:
Is there any better way to test out all scenarios. Have you tested this PR with all edge cases?

soumyamahunt:
> Is there any better way to test out all scenarios. Have you tested this PR with all edge cases?

I don't know if there is anyway we can test prelaunch scenario. The only reference we have how app would behave under prelaunch is mentioned in documentation but I don't know how to test it. As for normal launch/file activation/command line activation everything works as expected.

JasonStein:
@soumyamahunt  I made some changes and pushed directly to your remote branch, can you have a look?

soumyamahunt:
> @soumyamahunt I made some changes and pushed directly to your remote branch, can you have a look?

LGTM. 
> Also the naming of "OnSessionBackupAndRestoreOptionForInstanceChanged" is not clear. Maybe name it InstanceTypeChanged?

There is already an event named `OnInstanceTypeChanged` in `App.xaml.cs`, also `OnSessionBackupAndRestoreOptionForInstanceChanged` only invoked if the session snapshot settings value changes after instance type initialization is complete.

PR #830, opened 2021-03-25
Modify main GitHub action to include other configurations and auto-generated changelog

soumyamahunt:
1. This PR makes the `Notepads CI/CD Pipeline` action to run for other configurations like old `build` action used to do. The conditions are adjusted so Sonarcloud only runs with `Production` configuration, while for both `Debug` and `Release` config only build is tested.
2. Added auto generated changelog.
3. Introduced ignoring execution with documentation or other unnecessary changes.

## PR Type

Feature


aleks-ivanov:
@JasonStein @soumyamahunt these are all very beneficial improvements to the current versioning process, but before merging please wait for the PR with the automatic versioning for comparison, because the steps concerning the versioning are removed and replaced by a completely different implementation

soumyamahunt:
> @JasonStein @soumyamahunt these are all very beneficial improvements to the current versioning process, but before merging please wait for the PR with the automatic versioning for comparison, because the steps concerning the versioning are removed and replaced by a completely different implementation

I have made this PR draft and will retarget it to the new automatic versioning PR once that becomes live.

JasonStein:
Do you know why these steps are getting stuck instead of skipped?
image

.github/workflows/main.yml
```diff
@@ -50,26 +108,47 @@ jobs:
...
+          $LATEST_VERSION = [System.Version]::Parse($LATEST_TAG -replace 'v')
+          echo "::set-output name=semver::$(echo "$($LATEST_VERSION.Major).$($LATEST_VERSION.Minor).$($LATEST_VERSION.Build)")"
+        env:
+          GIT_REPOSITORY: ${{ github.repository }}
+
+      - if: matrix.release && steps.check_latest_tag.outputs.semver != ''
         name: Bump GitHub tag
         id: tag_generator
-        uses: mathieudutour/github-tag-action@v5
+        uses: soumyamahunt/github-tag-action@test-other-ver-support
```
JasonStein, line 125: Is this your own version of tag action?

JasonStein (changes requested):

.github/workflows/main.yml
soumyamahunt: I was about to comment about this, I opened a [PR](https://github.com/mathieudutour/github-tag-action/pull/73) in the action repo that will allow notepads to use the tag action without having to follow semver style tags. Currently I am keeping this for testing, but if the PR accepted we will follow the main repo.

soumyamahunt:
@JasonStein I have some suggestion about release strategy. Right now any push commit with the proper format triggers a release, the drawback of this is the changelog will only contain that specific commit and you will have to later modify changelog. What I want to propose is trigger release with `workflow_dispatch` event with an [input](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#example-workflow-configuration), this way we can always follow the format for all commits and the changelog will be generated from appropriate commits without requiring you to modify anything later.

.github/workflows/main.yml
JasonStein: > @JasonStein I have some suggestion about release strategy. Right now any push commit with the proper format triggers a release, the drawback of this is the changelog will only contain that specific commit and you will have to later modify changelog. What I want to propose is trigger release with `workflow_dispatch` event with an [input](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#example-workflow-configuration), this way we can always follow the format for all commits and the changelog will be generated from appropriate commits without requiring you to modify anything later.

100% agreed.

soumyamahunt:
@JasonStein I have changed the release process, have a look on the documentations and let me know if I should change anything.

PR #831, opened 2021-03-25
Switch to automatic versioning

aleks-ivanov:
### Here is how the automatic versioning works:

- a prerequisite is that the latest tag in the project's repo, should be converted from Assembly Versioning `1.4.3.0` to the industry standard [Semantic Versioning](https://semver.org/) `1.4.3`, because the step `name: Bump GitHub tag` that handles the automatic versioning is looking specifically for a Semantic Versioning tag (this change in the GitHub tagging **won't affect the project's nor package versioning, which is kept to Assembly Versioning in the `Package.appxmanifest`**)
- as mentioned in the previous PR, to activate the auto versioning the commit message should contain one of these three labels:
 
  - **perf:** + **BREAKING CHANGE:** - for major
  - **feat:** - for minor
  - **fix:** - for patch
 
- in the ci-cd documentation, you can find a detailed explanation of the commit process with examples
- when a commit to the default branch with a label in the message is pushed, the following happens:
  - a new GitHub tag is bumped according to the label and pushed
  - the `Identity.Version` value is replaced by the new version WITH a `.0` at the end, so that the Assembly Versioning in the `Package.appxmanifest` is preserved
  - down the pipeline everything executes the same as before
- the CD job executes only when a commit to the default branch with a label in the message is pushed

### There are a few overall changes:

- `strategy.matrix` for build configuration replaced simply by `"Production"` value
- `name: Install .NET Core` step removed, because it is unnecessary
- `name: Lowercase string generator` step removed and the `$LOWERCASE_REPOSITORY_OWNER = "${{ github.repository_owner }}".ToLower()` is added directly to the `name: Initialize SonarCloud scanner` step

@JasonStein @soumyamahunt  Please let us know if there any questions or concerns with this implementation 🙂

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,17 +5,14 @@ on: [push, pull_request, workflow_dispatch]
 jobs:
   ci:
     runs-on: windows-latest
-    strategy:
-      matrix:
-        configuration: [ Production ]
     outputs:
-      new_version: ${{ steps.get_assembly_version.outputs.version_num }}
-      new_version_tag: ${{ steps.get_assembly_version.outputs.version_tag }}
-      latest_tag: ${{ steps.get_latest_tag.outputs.tag }}
+      new_version: ${{ steps.tag_generator.outputs.new_version }}
+      new_version_tag: ${{ steps.tag_generator.outputs.new_tag }}
       is_push_to_master: ${{ steps.step_conditionals_handler.outputs.is_push_to_master }}
     env:
       SOLUTION_NAME: src\Notepads.sln
-      CONFIGURATION: ${{ matrix.configuration }}
+      CONFIGURATION: Production
+      DEFAULT_DIR: ${{ github.workspace }}
     steps:
       - name: Steps' conditionals handler
         id: step_conditionals_handler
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -59,39 +51,23 @@ jobs:
...
-        name: Get latest tag
-        id: get_latest_tag
-        shell: pwsh
-        run: |
-          $LATEST_TAG = git -c 'versionsort.suffix=-' ls-remote --exit-code --refs --sort='version:refname' --tags "https://github.com/$env:GIT_URL.git" '*.*.*' | tail --lines=1 | cut --delimiter='/' --fields=3
-          echo "::set-output name=tag::$(echo $LATEST_TAG)"
+          $APPXMANIFEST_PATH = 'src/Notepads/Package.appxmanifest'
+          $xml = [xml](Get-Content $APPXMANIFEST_PATH)
+          $xml.Package.Identity.SetAttribute('Version', "$env:NEW_VERSION.0")
+          $xml.save($APPXMANIFEST_PATH)
```
soumyamahunt, line 68: @aleks-ivanov This step will not change the package version in the manifest file in repo, right?? We have to manually change the original manifest to keep things in sync??

aleks-ivanov: @soumyamahunt it **does** change it, because the new version as I understand is needed for the packaging process.

aleks-ivanov: @soumyamahunt would like the `Package.appxmanifest` file to be committed to the repo with the updated version ?

soumyamahunt: > @soumyamahunt would you like the `Package.appxmanifest` file to be committed to the repo with the updated version ?

It's not necessary since we are using previous tag to genarate new version regardless of what manifest file version number. Ignore my previous comment.

PR #834, opened 2021-03-28
Adding quick button

korenevskiy:
Adding quick button functionality to the title bar.
99% completed.
Help is needed.
img-2021-03-28-06-17-01
img-2021-03-28-06-16-03
Dear friends and colleagues.
I have Visual Studio buggy. I can't make a Pool request.
I had to use "GitHub Desktop".
First I made the buttons in Visual Studio. But then there were problems.
I used notepad to make separate changes to the project folder, I used "GitHub Desktop" to make a Pull request.
I ask for help to conduct a check and can fix some errors.

### DIFF
--- a/src/Notepads/Services/AppSettingsService.cs
+++ b/src/Notepads/Services/AppSettingsService.cs
@@ -23,6 +23,50 @@
         public static event EventHandler<bool> OnSessionBackupAndRestoreOptionForInstanceChanged;
         public static event EventHandler<bool> OnSessionBackupAndRestoreOptionChanged;
         public static event EventHandler<bool> OnHighlightMisspelledWordsChanged;
+        public static event EventHandler<bool> OnQuickButtonsChanged;
+
+        private static List<string> _topQuickButtons = new List<string>();
+
+        public static List<string> TopQuickButtons
+        {
+            get => _topQuickButtons;
+            set
+            {
+                var isAdd = _topQuickButtons.Count() < value.Count(); 
+
+                
+
+                _topQuickButtons = value.Distinct().ToList();
+                OnQuickButtonsChanged?.Invoke(null, isAdd);
+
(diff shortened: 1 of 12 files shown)


JasonStein:
The reason I am not adding the button to the right area of the window is that there is a reserved part of the title bar which is not allowed to be used by the application. Windows specifically reserved that part of the title bar for the positioning function to work. This makes any kind of implementation on the right area of the title bar not work in the design perspective so I choose to not do it.

korenevskiy:
> The reason I am not adding the button to the right area of the window is that there is a reserved part of the title bar which is not allowed to be used by the application. Windows specifically reserved that part of the title bar for the positioning function to work. This makes any kind of implementation on the right area of the title bar not work in the design perspective so I choose to not do it. Thank you for the PP anyway but next time, you can use Github Discussion function first, and let's brainstorm the feature together before starting the work.

Now it is clear that the area on the right is not available.
These buttons are optional. The user may or may not enable them.
.
But if the area on the right is bounded. We can develop a function to choose the location of the buttons to the right of the tabs and the location to the left of the tabs..
In any case, this is an optional mode.
.
It's important for me to use notepad as notes, so it's important for me to have a recent open files button. My wife prints a lot, it is important for her to have a print button.
Therefore, we turn off these buttons. If it was the same for everyone, then there is no point in making the buttons temporary, they are made permanent. I made them disabled, so they do not interfere but complement.
.

korenevskiy:
> The reason I am not adding the button to the right area of the window is that there is a reserved part of the title bar which is not allowed to be used by the application. Windows specifically reserved that part of the title bar for the positioning function to work. This makes any kind of implementation on the right area of the title bar not work in the design perspective so I choose to not do it. Thank you for the PP anyway but next time, you can use Github Discussion function first, and let's brainstorm the feature together before starting the work.

**I suggested this idea in the discord chat.
Soumya answered me and told me to do PR.**

korenevskiy:
@JasonStein 
img-2021-03-28-23-37-11

PR #842, opened 2021-03-29
ci: Ignore more unnecessary events for CI and Dependabot improvements

soumyamahunt:
1. PRs submitted by bots trigger both `push` and `pull_request` events, since bots don't have access to some secrets CI in `push` event fails. This PR fixes that by ignoring `push` event in those cases.
2. Dependabot now follows the new PR title convention.
3. Also, CI doesn't run when tags are pushed as tag generation is automated and doesn't require any tests.

## PR Type

- CI/CD pipeline changes

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,17 +5,23 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    commit-message:
+      prefix: 'github-action-deps: '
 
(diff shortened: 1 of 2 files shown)


JasonStein:
@soumyamahunt 
image

soumyamahunt:
> @soumyamahunt
> image

Fixed in #843.

PR #849, opened 2021-04-01
ci: Use path matching to dismiss scanned alerts

soumyamahunt:
This PR improves the bulk dismissal task by allowing to provide paths to ignore. The task asks for input for choosing filter to run, providing `desc` dismisses alerts based on description, providing `path` dismisses alerts based on path (default is `path`).

## PR Type

 - CI/CD pipeline changes

### DIFF
--- a/.github/workflows/csa-bulk-dismissal.yml
+++ b/.github/workflows/csa-bulk-dismissal.yml
@@ -59,26 +59,21 @@ jobs:
         shell: pwsh
         run: |
           $page = 1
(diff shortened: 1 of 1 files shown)


soumyamahunt:
@JasonStein currently the dismiss reason is provided as a secret, I think it would be more accessible if it can be provided as input when triggering action. What are your thoughts??

JasonStein:
> @JasonStein currently the dismiss reason is provided as a secret, I think it would be more accessible if it can be provided as input when triggering action. What are your thoughts??

For now, we can put an arbitrary reason for the dismisal. Don't spend too much time on this can we can revisit this later once we have some useful insights by the scan.

PR #852, opened 2021-04-02
ci: Upload app payload with dependencies as zip in Release

soumyamahunt:
Rename the .msixbundle release asset to the display name

## PR Type

- CI/CD pipeline changes

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -338,12 +338,21 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Rename msixbundle
+        id: rename_notepads_zip
+        shell: pwsh
+        run: |
+          Rename-Item $env:ASSET_PATH $env:NEW_NAME
+        env:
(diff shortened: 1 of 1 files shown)


soumyamahunt:
@JasonStein another thing I would like to suggest is to upload whole folder that contains dependencies as zip so users having issue with MS store can install the app without having to rely on store to download dependencies.

JasonStein:
> @JasonStein another thing I would like to suggest is to upload whole folder that contains dependencies as zip so users having issue with MS store can install the app without having to rely on store to download dependencies.

That will increase the size of the package significantly right?

soumyamahunt:
> > @JasonStein another thing I would like to suggest is to upload whole folder that contains dependencies as zip so users having issue with MS store can install the app without having to rely on store to download dependencies.
> 
> That will increase the size of the package significantly right?

This provides both .msixbundle and .zip in releases. So users can download bundle file which will be of the same size as before and install it. If they have any issue with installing via bundle file they can download the zip file (which will be quite larger than bundle file) and install Notepads by executing `Install.ps1`.

PR #854, opened 2021-04-07
ci: Fix error in nuget package restore

soumyamahunt:
Fix action error on nuget package restore step.

## PR Type

- CI/CD pipeline changes


soumyamahunt:
@JasonStein Restoring via `nuget.exe` succeeds while `msbuild` fails. Looks like MS screwed up something in latest `msbuild` release. I have [reported](https://github.com/microsoft/setup-msbuild/issues/47) this issue.

PR #857, opened 2021-04-08
fix: Fix pre-launched instance couldn't be promoted to primary instance from shadow instance

soumyamahunt:
Fix pre-launched instance couldn't be promoted to primary instance from shadow instance.

## PR Type

- Bugfix

### DIFF
--- a/src/Notepads/Views/MainPage/NotepadsMainPage.MainMenu.cs
+++ b/src/Notepads/Views/MainPage/NotepadsMainPage.MainMenu.cs
@@ -9,7 +9,7 @@
     using Windows.UI.Xaml;
     using Windows.UI.Xaml.Controls;
     using Windows.UI.Xaml.Controls.Primitives;
-    using Windows.UI.Xaml.Media;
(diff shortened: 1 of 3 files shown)


soumyamahunt:
@JasonStein pre launching can only be tested when app is packaged so you have to package it first and test it yourself or provide the package to me to test. Here is the different test scenario:

1. Disable `"Always open new window"` in settings and launch the app (The app will register itself for pre-launch), check everything in the UI reflects the app is primary instance (main button color, shadow indicator in status bar present or not, settings button enabled, session snapshot enabled etc.).
2. Then wait for app to be pre-launched (We have no control over this), you can check if Notepads is pre-launched or not by going into task manager and see if any suspended Notepads process exist in `Background process` section.
3. Launch app again if Step-2 is success and check if app is primary instance or not as in Step-1.

Test the above steps with session snapshot enabled to check if app restores session properly. Although `"Always open new window"` enabled scenario works, still you can test for that too.

PR #866, opened 2021-04-12
fix: fix session snapshot not being properly loaded

soumyamahunt:
Handled scenarios that could have prevented session snapshot not being properly loaded.

## PR Type

- Bugfix

### DIFF
--- a/src/Notepads/App.xaml.cs
+++ b/src/Notepads/App.xaml.cs
@@ -28,13 +28,28 @@
 
         public static Guid Id { get; } = Guid.NewGuid();
 
-        public static bool IsPrimaryInstance = false;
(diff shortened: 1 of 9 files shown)


soumyamahunt:
@JasonStein I would suggest to not merge this until we have properly tested the fix works.

soumyamahunt:
@JasonStein I found a way to [debug prelaunch scenario from VS](https://devblogs.microsoft.com/devops/debug-uwp-prelaunch-with-vs2015/) and tested and sessions loaded correctly. If user opens any file that already present in sessions then the file will be removed and the sessions will be loaded. You can test this with the procedure described in blog post.

PR #870, opened 2021-04-17
fix: fix command line activation not working properly in some cases

soumyamahunt:
* Fixed command line activation not working for debug build of app.
* Implemented unit tests for testing command line functionality.
* Added checks in CI/CD pipeline for unit tests.

## PR Type

- Bugfix

- CI/CD pipeline changes


soumyamahunt:
@JasonStein trying to open files with debug mode in notepads in cmd was failing, I made changes to fix such issues and also added all the test cases I could think of for command line activation. Check them if you can think of some more could be added or some test cases contain invalid result comparison. I added checks in CI so that pipeline will fail if tests aren't passed.

Also, I added `removableStorage` permission to manifest, this permission would give Notepads  access to external storage so that launching files from command line will be more successful. This permission is given silently without any prompt to user.

PR #872, opened 2021-04-20
fix: update flow direction properly when changing language in app context

### LINKED ISSUE #773: [Bug] Tabs position in RTL locales
Thank you for the awesome notepad.

**Describe the bug**
Tabs are mispositioned in RTL locales 

**To Reproduce**
Change language to the newly added "العربية (اليمن) [#772] , then restart Notepads

**Expected behavior**
Close & Maximize & Minimize buttons should be appear on the left and the tabs on the right ideally

**Screenshots**
image

Tabs should be beside the Close & Maximize & Minimize box or ideally on the right.

also the controls and texts align should be RTL too
Settings

**Desktop:**
 - OS: [e.g. Windows 10 20H2 19042.746]
 - Version [e.g. v1.4.2.0]

soumyamahunt:
Fixes #773.

## PR Type

- Bugfix

### DIFF
--- a/src/Notepads/App.xaml.cs
+++ b/src/Notepads/App.xaml.cs
@@ -16,6 +16,7 @@
     using Windows.ApplicationModel.Activation;
     using Windows.ApplicationModel.Core;
     using Windows.ApplicationModel.DataTransfer;
+    using Windows.ApplicationModel.Resources.Core;
     using Windows.UI;
     using Windows.UI.ViewManagement;
     using Windows.UI.Xaml;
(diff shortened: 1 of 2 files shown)


src/Notepads/App.xaml.cs
```diff
@@ -168,6 +169,16 @@ private async Task ActivateAsync(IActivatedEventArgs e)
         private Frame CreateRootFrame(IActivatedEventArgs e)
         {
             Frame rootFrame = new Frame();
+
+            var flowDirectionSetting = ResourceContext.GetForCurrentView().QualifierValues["LayoutDirection"];
```
JasonStein, line 173: This seems hacky, are you sure this is the correct or proper way of detecting RTL?

soumyamahunt: Yes, I found it from [documentation](https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.resources.core.resourcecontext.QualifierValues?view=winrt-19041) and tested it. This is the only way we can fix this bug until MS [fixes](https://github.com/microsoft/microsoft-ui-xaml/issues/4891) it themselves.

PR #976, opened 2021-10-30
Update to Windows 11 style, WinUI 3.0

BobbyESP:
## PR Type

- Updated most of the components of the app to the Windows 11 style, WinUI 3.0

### Bugs known
- Settings NavigationView doesn't load the first NavigationItem

Captura de pantalla 2021-10-30 182615

- Some strange errors when opening the app in App.xaml.cs file, line 50 related with AppCenter (never heard about that)

### DIFF
--- a/src/Notepads/App.xaml
+++ b/src/Notepads/App.xaml
@@ -1,12 +1,17 @@
 ﻿<Application
     x:Class="Notepads.App"
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
-    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:controls="using:Microsoft.UI.Xaml.Controls">
 
     <Application.Resources>
-        <ResourceDictionary>
 
-            <!--<ResourceDictionary.ThemeDictionaries>
+
+        <controls:XamlControlsResources>
+            <controls:XamlControlsResources.MergedDictionaries>
+                <ResourceDictionary>
+                    
+                    <!--<ResourceDictionary.ThemeDictionaries>
                 <ResourceDictionary x:Key="Light">
 
                 </ResourceDictionary>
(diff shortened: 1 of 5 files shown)


BobbyESP:
The errors in App.xaml.cs:
Captura de pantalla 2021-10-30 183705
Captura de pantalla 2021-10-30 183749
Captura de pantalla 2021-10-30 183730

BobbyESP:
Ah yes, forgot to tell that the Navigation Item that doesn't open when you open settings can be opened by clicking on it. Just fails when you open the settings.

JasonStein:
I will be working on the visual update by EOY. I will wait for WinUI 2.x to get mature before doing so.

BobbyESP:
It's ok. If you need some help, tell me. WinUI 3 by the moment isn't avaliable for UWP or it's very unestable, so better wait.

onein528:
Delete all custom style and update UI:
image

onein528:
@BobbyESP Hai. how have you seen:)

> Settings NavigationView doesn't load the first NavigationItem

FYI, It's a not bug but specification. so, need to load the page in NaviView_Loaded() or Page_Loaded()

Instead, use the IsSelected flag to receive change events on ItemSelected. but this app uses ItemInvoked() this is not trigger selection changing event from IsSelected = true.

https://github.com/JasonStein/Notepads/blob/master/src/Notepads/Views/Settings/SettingsPage.xaml#L18

BobbyESP:
Ohh, i see!

JasonStein:
I am waiting for WinUI v2.8 stable release before looking into upgrading the UX. Are there any issues you find with v2.8 prerelease?

onein528:
ah. I don't have the problem I saw in WinUI prerelease in UI Control, but I found a lot in WerbView2.

Also, due to that issue, projects that download WinUI 2.8 prerelease may fail in store submissions.

onein528:
And when updating to WinUI 2.7 or later, the IsSelect property does not trigger the ItemInvoked event. It still triggers SelectionChanged, so you need to navigate to the initial page when loading the settings page. You will find it during the migration.

JasonStein:
> 

There are some other UI issues bringing by the v2.7 release, I cannot remmeber the exact but it has been disscussed earlier. The roll of the WinUI 2 and 3 releases are much slower than my expectation:(

KParthSingh:
Guys It has been over two years now. When will this UI update take place?

0x5bfa:
The owner hesitates to move to WinUI2.8 or 3.0 because it would bring more WinUI-side issues.

KParthSingh:
Damn.

emvaized:
> WinUI-side issues

Are these issues are still actual as for today?

0x5bfa:
I said them to the same thing. lol.

Issue #1046, opened 2022-02-25
[Feature request]


bearoxo:
**Is your feature request related to a problem? Please describe.**

- Notepads new file/tab can only default to .txt file, you'd need to save it as .md first before you can use markdown.

**Describe the solution you'd like**

- Since Notepads goal is to be a modern replacement for the old archaic Windows Notepad, perhaps using a more modern file extension like markdown .md as the default might be more suited for it, or maybe give users the option to choose their preferred default extension whenever they create a new file/tab.

**Describe alternatives you've considered**

- I just save the file as .md, but sometimes I just need to write something in markdown without really having the need to save it locally on my pc.

JasonStein:
I was planning to add this feature initially but decided to not do so because:

1. The Markdown editing feature Notepads currently provided is very bare minimum. It does not support all markdown syntax; it does not show local image (restriction of UWP) and performance is also not ideal. That's also the reason that I am hiding the feature by default.
2. I don't want to market the app as a Markdown editor, and I want to keep it as a plain text editor as much as possible.

PR #1077, opened 2022-03-28
feat: Make default file extension changeable

RLGingerBiscuit:
Addresses #1046, which requested the changing of the default file extension to .md, which I added along with the ability to change the default file extension in the Advanced settings tab.

image

## PR Type

- Feature

## Other information

It's probably not the best, however I wanted to change it for myself and decided to open a PR for it.
I only speak English, but I have added the needed options in all the Settings.resw files, along with removing the file extension from Resources.resw. I tried to adhere to the code already present, however a couple things may need changing as I am not used to UWP.


D-Maxwell:
This does not appear to have been implemented, unless the setting's location was changed.
Are there any plans to bring this back ?
Not implying I'd require a markdown editor by any means, yet I wish I could set the default file extension of the output file.

PR #1194, opened 2023-02-04
vi-VN: Update to Vietnamese

quanultra:
## PR Type
- Translation


0x7c13:
The length of Settings.resx is not matching other lans. Can you double check all files have same length with other translations?

quanultra:
@0x7c13 
i was update the Settings.resx file, please take a look.

PR #1324, opened 2024-03-16
style: improvements to a couple of UserControls

United600:
Improved the look and feel of the 'DismissButtonStyle' and the respective UserControls that make use of it.

## PR Type

- Other... Please describe:

Better padding, sizes, glyphs and brushes.

## Other information

Added some design attributes.


0x7c13:
@United600 I have played around with latest WinUI3/WinAppSdk and decided to adopt the design and app model once its get matured. So far, there are still lots of issues presented（I have spent 30+ hours building an app using WinAppSDK so I know what I am speaking. NativeAOT support is also just realized by most of the parties. Guess we still need to wait for some time before taking any major actions. FYI: Your changes will be obsolete once we move to settings card controls etc when adopting WinUI3.

United600:
That's sad to hear. Personally, I would have opted for WinuUI 2 first and stuck with UWP in the meantime.
Sure some of the changes here would be obsolete, eg. Expander, however they are applicable to a future migration.

